Issue #140, opened 2021-03-04
likelihood values for some quantities with lambda = 0 not getting reported


iantaylor-NOAA:
Vlada Gertseva and I noticed inconsistent reporting of the likelihoods for values with lambda = 0.

The mean size at age likelihood is only calculated when the lambda > 0 controlled by this line:
https://github.com/nmfs-stock-synthesis/stock-synthesis/blob/e28f64dc3a4eb00f842fcdddd03d3b61ccfa1d18/SS_objfunc.tpl#L458

In contrast, other components also do the calculation if save_for_report > 0, as below:
https://github.com/nmfs-stock-synthesis/stock-synthesis/blob/e28f64dc3a4eb00f842fcdddd03d3b61ccfa1d18/SS_objfunc.tpl#L157

I can push this change if you want, but don't want to get in the way of ongoing work in other branches. I also haven't done a comprehensive search of which other likelihood components are missing the `|| save_for_report>0`.

Rick-Methot-NOAA:
Go for it.  I am far away from that part of the code.
get the issue included in the new project tab

iantaylor-NOAA:
Issue resolved in Pull Request #141. 
I've changed the Issue Prioritization to "Done" and will close this issue. 
It is listed on the [3.30.17 release issues](https://github.com/nmfs-stock-synthesis/stock-synthesis/milestone/1) Milestone so can be noted on the change log if appropriate.

PR #141, opened 2021-03-05
fix: always calculate likelihoods when save_for_report>0 #140

iantaylor-NOAA:
This is an attempted fix for issue #140. I tested locally on the Big Skate model which has lambda = 0 for mean size at age data and the likelihood is now reported (without contributing to the total likelihood).

I also found 4 other locations where adding `save_for_report>0` seemed necessary for consistent reporting, a couple of which already had more complex if() statements.

@Rick-Methot-NOAA, could you check the change in this pull request to confirm that I put the parentheses and || in the right places before merging this change?

### DIFF
--- a/SS_objfunc.tpl
+++ b/SS_objfunc.tpl
@@ -455,7 +455,7 @@ FUNCTION void evaluate_the_objective_function()
     if(nobs_ms_tot>0)
     {
       for (f=1;f<=Nfleet;f++)
-      if(Nobs_ms(f)>0 && sizeage_lambda(f,k_phase)>0.0)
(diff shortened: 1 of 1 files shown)


Rick-Methot-NOAA (approved):
code changes look good.  Possible that this could disclose some uninitialized variable, so examine  results closely.

Rick-Methot-NOAA:
Great.  We have a good system in place.


*Richard D. Methot Jr. Ph.D.**NOAA Fisheries Senior Scientist** for Stock
Assessments*

*Mobile: 301-787-0241*

PR #160, opened 2021-04-13
Issue#154 smrybio

Rick-Methot-NOAA:
test ready for improvements to totbio and smrybio reporting

### DIFF
--- a/SS_biofxn.tpl
+++ b/SS_biofxn.tpl
@@ -477,10 +477,12 @@ FUNCTION void get_growth2(const int y)
             if(s==nseas) add_age=1; else add_age=0;   //      advance age or not
 // growth to next season
             VBK_temp2=(mfexp(VBK_temp*seasdur(s)*VBK_seas(s))-1.0);
+//    warning<<t<<" ave_size_grow2_start "<<Ave_Size(t,1,1)(0,6)<<endl;
             switch (Grow_type)
             {
               case 1:
               {
+              	Ave_Size(t+1,1,g,0)=Ave_Size(t,1,g,0);  // carryforward, but may be overwritten in some circumstances
                 for (a=0;a<=nages;a++)
                 {
                   if(a==nages) {k2=a;} else {k2=a+add_age;}  // where add_age =1 if s=nseas, else 0  (k2 assignment could be in a matrix so not recalculated
@@ -506,8 +508,9 @@ FUNCTION void get_growth2(const int y)
                   }
                   else  // in linear phase
                   {
-                    Ave_Size(t+1,1,g,a) = len_bins(1)+lin_grow(g,ALK_idx,a)*(Cohort_Lmin(gp,y,a)-len_bins(1));
+                    Ave_Size(t+1,1,g,k2) = len_bins(1)+lin_grow(g,ALK_idx,a)*(Cohort_Lmin(gp,y,a)-len_bins(1));
                   }
+//                  if(y==1990 && g==1) warning<<a<<" "<<lin_grow(g,ALK_idx,a)<<" "<<Ave_Size(t,1,g,a)<<" "<<Ave_Size(t+1,1,g,k2)<<endl;
                 }  // done ageloop
                 break;
               }
@@ -651,6 +654,9 @@ FUNCTION void get_growth2(const int y)
             if(do_once==1) echoinput<<"  seas: "<<s<<"  size@t+1:  "<<Ave_Size(t+1,1,g)(0,min(6,nages))<<" plusgroup: "<<Ave_Size(t+1,1,g,nages)<<endl;
  #endif
            }  // end of season
+
+ /*
+//  move this code to popdyn in styr so can use adjustments made by growth3
 //  SS_Label_Info_16.2.4.3  #propagate Ave_Size from early years forward until first year that has time-vary growth
           k=y+1;
           j=yz+1;
@@ -668,6 +674,7 @@ FUNCTION void get_growth2(const int y)
             k++;
             if(j<endyr+1) j++;
           }
+ */
         }  // end need to consider this GP x settlement combo (usemorph>0)
       }  // end loop of settlements
       Ip+=N_M_Grow_parms;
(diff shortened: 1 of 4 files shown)


k-doering-NOAA:
Great, @Rick-Methot-NOAA ! I will see if the smrybio and total values are now matching up in ss_summary.sso and Report.sso - anything else that you would like me to check?

Rick-Methot-NOAA:
Please do the check that you mention and then merge the pull request.

k-doering-NOAA:
I'm still seeing differences between the total bio during model years (forecast years are fine) reported in ss_summary.sso and Report.sso. As an example, in 1971 for simple:
ss_summary_report_mismatch

I am also attaching the ss_summary.sso and Report.sso in a zipped file: 
[simple_output_files.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/6305104/simple_output_files.zip)

Rick-Methot-NOAA:
smrybio is OK, but I also see a diff for totbio.  So it must be in the
recruits.  Will check tonight.

k-doering-NOAA:
> smrybio is OK, but I also see a diff for totbio. So it must be in the recruits. Will check tonight.

Thanks! Yes, that is exactly what I am seeing as well, smrybio values are fine. For totbio, the forecast years, init, and virgin values are fine, too. It is just the years within the model that are not matching.

Rick-Methot-NOAA:
I know what the issue is.  Should be easy fix, but a bit of a conceptual
dilemma as to whether age 0 fish that are just recruiting should be part of
totbio.

iantaylor-NOAA:
I like having age-0 fish part of totbio. For most stocks their contribution is pretty trivial, but it makes it easier to explain what totbio is, and smrybio is available for anyone who wants only 1+.

Rick-Methot-NOAA:
The problem is that smrybio and totbio are calculated and stored at the top
of the year (before the season loop and before that year's recruitment can
happen).  In some model configurations, recruitment does not even happen on
Jan 1.  These smrybio and totbio values are what is reported in ss_summary
and they are summed over all areas.  For models that have recruitment on
Jan 1, it would be easy to modify the so that the age 0  recruits get
included in the totbio.  But if recruitment is in a later season, then
complications arise.

Then in report.sso, the time_series table recalculates these quantities for
EACH area.  When it does so, it uses stored numbers-at-age, including age
0.  Hence the difference.

A further complication is with models that set age-at-recruitment greater
than age 0, so age 0 fish never exist.  There is no accommodation for this
in the reporting.

iantaylor-NOAA:
@Rick-Methot-NOAA, all that makes sense, thank you.
It makes me think that we shouldn't be using totbio at all and should just rely on smrybio. However, I suppose removing it at this point would be more trouble than just working out an adequate solution to the differences.

k-doering-NOAA:
Wow, there is a lot of nuance here. @Rick-Methot-NOAA , is it possible to write the ss_summary.sso at the same time (or right after) the second call to create Report.sso? I think most people would expect the ss_summary file to be giving the same information as Report.sso, but just less. To me, it seems like ss_summary.sso should be using the exact same values as in Report.sso. Could that be a solution?

We could also add some information to the manual about how the total bio value is being calculated , perhaps in the section on the report file?

Rick-Methot-NOAA:
Seems like too much detail for the manual.  I will try to add some
explanatory description to the top of the report file where other aspects
of the model configuration are described.  It will include whether, or not,
totbio includes age zero according to the particular model configuration.

Rick-Methot-NOAA:
I now am thinking that we should do totbio as 1+ and clarify what that
includes.  If users want to see age 0, they can get it from the
age-specific tables.

Rick-Methot-NOAA:
Kathryn,
Unless your sharper eyes see something that I have missed, I am ready to merge this into main.

k-doering-NOAA:
@Rick-Methot-NOAA taking a look now!

k-doering-NOAA:
Great work @Rick-Methot-NOAA , I don't see any reason to not merge this into main, so I think you can do so!

A few things I checked
- Total Bio and Summary Bio matching between summary and report files for simple model
- Total Bio and Summary Bio matching between summary and report files for three area no movement model

I also ran jenkins with no estimation (just b/c the github no est workflow wasn't available on this branch)  and found no large enough differences for it to flag. The models are running with estimation currently, but I'm not sure if we will want to wait for these or not (won't be done for ~ 8 hrs)

I did note one small difference on the jenkins no estimation run, but this also appears on main (with the exact difference), so it is not due to the work in this branch and I don't think there is any reason to not merge this work into main.

The forecast catch value changed slightly from the reference:

| "quantity" | "value" | "ref_value" | "ratio" | "mod_name" |
| ---------- |  -------- | ------------ | ------- | -------------- |
| "ForeCatch_2027_val  |     " " 2.95209e+03" | " 2.94694e+03" | " 1.001748" | "vermillion_snapper" |

k-doering-NOAA:
One last comment about documentation (this is just my opinion):  I still think it might be helpful to include information about what these values are in the manual (even if it is just a sentence or 2). 

I know we have been hesitant to expand the manual, but I think because the goal of the manual is to be a reference document, the more details, the better (as long as they are relevant to the use of SS by users). How else would users know for instance that Total bio doesn't include age 0 fish by default? My assumptions would surely be it is included. My impression based on the users I have worked with is that many people rely heavily on the manual for details on how features work. If it is not there, the feature probably doesn't get used or the SS dev team will have to explain how it works (for users brave enough to ask), which takes time.

A blog I read recently that got me thinking a little differently about documentation: https://third-bit.com/2019/04/16/what-docs-when/  .

Rick-Methot-NOAA:
OK.  In the end I did keep the code including age 0 in totbio because most
model configurations put all recruits on Jan 1.

I think the only clarification we need is that totbio and smrybio are both
calculated on Jan 1.  If the model configuration puts some recruits to
occur after Jan 1 or in a later season, then the totbio quantity may not
provide the information you want.  In that case, consult the detailed age-,
area-, and season-specific tables in report.sso.

k-doering-NOAA:
> OK. In the end I did keep the code including age 0 in totbio because most model configurations put all recruits on Jan 1.

> I think the only clarification we need is that totbio and smrybio are both calculated on Jan 1. If the model configuration puts some recruits to occur after Jan 1 or in a later season, then the totbio quantity may not provide the information you want. In that case, consult the detailed age-, area-, and season-specific tables in report.sso.

Yes, that makes sense to me, thanks for clarifying. I agree including a sentence like the one above seems sufficient.

Issue #157, opened 2021-04-08
set up SS no estimation job on gh actions


k-doering-NOAA:
Also, we could consider if we want the running with no estimation job to run on github actions instead of on jenkins. The benefit of this is we could set it up to run automatically upon pushes to SS , which is not possible with the way the Jenkins server is set up (only can manually start or run on schedule).

_Originally posted by @k-doering-NOAA in https://github.com/nmfs-stock-synthesis/ss-test-models/issues/1#issuecomment-791653480_

k-doering-NOAA:
This is in progress in the gh_actions_no_est branch.  So far, the environment has been set up, with a compiled version of SS, the model input/output files, and R available to use. 

Next steps are to generalize or modify the code from jenkins so that it will run on gh actions. Another solution would be to just move everything around so it has the same relative path as on jenkins, but there are likely side benefits to generalizing (in case we need to move to a new system in the future, for exampe).

k-doering-NOAA:
Merged finished job into main, so closing

PR #161, opened 2021-04-13
testing, issue #157 : set up no estimation model runs for each commit

k-doering-NOAA:
This adds a new check on every commit to SS (we can change this frequency if it seems like too much ) where the ss-test-models are run with no estimation and checked if the values are the same or different as in the reference model. Hopefully this will catch changes that impact calculations in the SS code base more quickly.

@Rick-Methot-NOAA , if the checks pass and this description of how this works makes sense to you, please feel free to merge this in! Note it will only be available in main after merging, so if we want to use it in other branches we will need to manually create it (which I can do).

### DIFF
--- a/.github/workflows/run-no-est.yml
+++ b/.github/workflows/run-no-est.yml
@@ -1,44 +1,75 @@
 # run models without estimation
 name: run-no-est
-on: [workflow_dispatch]
+on: [push, pull_request]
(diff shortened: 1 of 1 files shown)


Rick-Methot-NOAA:
I approve this change to the testing procedure.

Issue #158, opened 2021-04-09
SPR/YPR report not written to report file even if hessian is inverted


k-doering-NOAA:
I noticed that in several models (running from current  main branch), The `
 SPR/YPR_Profile report:54`  item in the control file is not written, but instead the statement `This output only included after the Hessian is inverted`  is shown, despite the fact the hessian was estimated for these models. I think this statement may not be working properly (perhaps wrote_bigreport not getting updated to be greater than 0?):
https://github.com/nmfs-stock-synthesis/stock-synthesis/blob/a923c2546dd9bff6aafd43c63bf2528a05d2f32e/SS_write_report.tpl#L3250-L3254

Rick-Methot-NOAA:
The logic hoop jumps I went through to get things not to report twice is the source of this problem.  I'll take a look.

Rick-Methot-NOAA:
A flag, SDmode, is created in readstarter and is 1 if Hessian is to be attempted, or 0 if -nohess is used.
This flag is now used to improve the logic of which reports to do at the end of the the run after sdreport is completed.

NOTE for future code improvement:
Global MSY and SPR/YPR profile currently are in write_report, but with some complicated logic
They could be moved into functions similar to write_Bzero_output and called from FINAL section in global.

Issue #159, opened 2021-04-09
warn users if they request se for fixed quantities


k-doering-NOAA:
Related to https://github.com/nmfs-stock-synthesis/ss-test-models/issues/3

@Rick-Methot-NOAA realized it would be useful to warn users if they request standard error vals for quantities that are fixed. This matters because it sometimes can be the only reason for an non positive definite hessian.

Rick-Methot-NOAA:
I think the MG_active() vector could be used to check to see if M or growth is estimated
  ivector MG_active(0,7)  // 0=all, 1=M, 2=growth 3=wtlen, 4=recr_dist&femfrac, 5=migration, 6=ageerror, 7=catchmult

there is no equivalent for the selparms, but there easily could be creation of a vector that was fleet-specific

Issue #154, opened 2021-04-05
discrepancies in smrybio and totbio reporting


Rick-Methot-NOAA:
Overall the calculation and storage of totbio and smrybio is messier and more convoluted than it should be.  Some are calculated on the fly, and some not until the write stage.  
The Time_Series table, which r4ss relies on, calculates then using:
     temp=natage(t,p,g)(Smry_Age,nages)*Save_Wt_Age(bio_t,g)(Smry_Age,nages);
     Bio_Comp(GP(g))+=value(temp);   //sums to accumulate across platoons and settlements
     Num_Comp(GP(g))+=value(sum(natage(t,p,g)(Smry_Age,nages)));
     totbio+= natage(t,p,g)*Save_Wt_Age(bio_t,g);
     smrybio+= temp;
     smrynum+=sum(natage(t,p,g)(Smry_Age,nages));
     smryage+=natage(t,p,g)(Smry_Age,nages)*r_ages(Smry_Age,nages);

But the values output for ss_summary.sso are calculated within ss_popdyn because smrybio needs to be available as a basis for density-dependent parameters.

Sometime after 3.30.16, the calculation within ss_popdyn went awry for purpose of smrybio reporting, but not for any other noticeable dependent purpose.

Also note that ss_summary was reporting smrybio for both smrybio and totbio for time_series years only; forecast years remained correct.

With this issue, the whole situation will be investigated and a consistent, tested approach implemented.

And note that

k-doering-NOAA:
Some observations using the Simple model, in case they are helpful. I ran simple using 3.30.16 and then using the [current main branch](https://github.com/nmfs-stock-synthesis/stock-synthesis/tree/a923c2546dd9bff6aafd43c63bf2528a05d2f32e).

In 3.30.16:
- Totbio values (labeled AllBio) and SmryBio for all years (model years and forecast years) are different in Report.sso, but are the same in ss_summary.sso. It seems like SmryBio values are being used for both quantities in ss_summary.sso. However, Totbio virgin and init in ss_summary.sso match Report.sso.
-  SmryBio_virgin is 0 in ss_summary.sso, but is 124144 in Report.sso.

In current main: (tried to italicize some observations to highlight the differences between .16 and main)
[...]

Issue #152, opened 2021-03-25
dirichlet multinomial parameters must be labeled consecutively


k-doering-NOAA:
It looks like Dirichlet multinomial parameters must be labeled consecutively from 1 to the last parameter in defining them in the data file in order for SS to read them correctly. For example, say a user has a model with DM parameters labelled 1 to 4 in the data file. If a user decides to drop a fleet, say, the one associated with DM parameter 1, then there will only be DM parameters for 2 to 4 (3 total parameters). SS will stil expect to read 4 parameter lines in the control file, though, and will exit on error when it cannot figure out which fleet parameter 1 (which is not longer being used)  belongs with. 

I think it would be helpful to either 1) exit on error and provide an informative msg when users have not used consecutive parameter numbers for DM in the data file or 2) revise the DM parameter option so that users do not need consecutive parameter numbers. The first is the simpler of the options, but I could see 2) being more useful.

Rick-Methot-NOAA:
I will take a look at this.  Please send me an example setup in which I can invoke the problem.

k-doering-NOAA:
Thanks, just forwarded you an email with an example.

Rick-Methot-NOAA:
The numbering of the requested D-M parameters must be consecutive and cannot exceed Nfleet.
I created a fatal warning if user tries to assign a D-M index > Nfleet.
I have not created an algorithm to figure out if any number gets skipped in the sequence.

So, this is legit and will take 6 parameters:
#_mintailcomp	addtocomp	combine_M_F	CompressBins	CompError	ParmSelect	minsamplesize
-0.001	1e-07	0	10	1	 1	0.01	#_f2_volpal  
-0.001	1e-07	0	10	1	 2	0.01	#_f3_trawlers
-0.001	1e-07	0	10	1	 3	0.01	#_f4_art     
-0.001	1e-07	0	10	1	 4	0.01	#_f5_cdtrw   
-0.001	1e-07	0	10	1	 4	0.01	#_SpSurv     
-0.001	1e-07	0	10	1	 5	0.01	#_CdSurv     
-0.001	1e-07	0	10	1	 5	0.01	#_PtSurv     
-0.001	1e-07	0	10	1	 5	0.01	#_SpCPUE     
-0.001	1e-07	0	10	1	6	0.01	#_PtCPUE     

[...]

PR #162, opened 2021-04-15
fixing issue #158  spr/ypr report

Rick-Methot-NOAA:
the logic for creating this report is fixed with this commit.  Check it out.

### DIFF
--- a/SS_global.tpl
+++ b/SS_global.tpl
@@ -691,7 +691,6 @@ FINAL_SECTION
     save_for_report=1;
     bigsaver=1;
     save_gparm=0;
-    wrote_bigreport=0;
     if(SDmode==0 && pick_report_use(60)=="Y") write_bodywt=1;  //  turn on conditional on SDMode because SDMode=1 situation already written
     y=styr;
     setup_recdevs();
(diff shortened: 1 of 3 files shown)


k-doering-NOAA:
@Rick-Methot-NOAA I tested with running with simple; when I run with hessian, I am now seeing the output "SPR/YPR profile deferred until after the Hessian is inverted" where the values should be. The switch is still not being triggered somehow... I suspect it may be `wrote_bigreport` not getting updated to 1 after the first write of the report file?

Rick-Methot-NOAA:
Well, I had it working.  Then I must have broken it again.  Maybe I will
make the bigger change I mentioned in order to make the whole logic train
more straightforward.

k-doering-NOAA:
Hmm, I tried running with -nohess and realized the YPR/SPR table did show up then! maybe the if/else statements are backwards?

Rick-Methot-NOAA:
Final fix involved moving SPR_profile and Global_MSY into separate functions (e.g. not part of write_bigoutput), so now they can be called separately as the final step before model finishes.  Could add a check so that they do not get done after a MCMC.

k-doering-NOAA:
I tested the new change locally and spr/ypr profile was written for both with and without hessian - is this the expected behavior, @Rick-Methot-NOAA ? 

Also, I see your note about adding a check for MCMC -  do we need to do any MCMC testing before merging this in?

Rick-Methot-NOAA:
That is the expected behavior, so all is good.

I see no need to add a test or code change regarding MCMC.  These three
report components are the final stages before the "run is complete"
message, so users can control via the custom report controls:
    if (pick_report_use(59)=="Y")    {
        cout<<"dynamic Bzero: ";
        write_Bzero_output();
        cout<<" finished "<<endl; }

    if (pick_report_use(54)=="Y" && Do_Benchmark>0)    {
        cout<<"SPR_profile: ";
        SPR_profile();
        cout<<" finished "<<endl;    }

    if (pick_report_use(49)=="Y" && Do_Benchmark>0)    {
        cout<<"Global_MSY: ";
        Global_MSY();
        cout<<" finished "<<endl;    }

k-doering-NOAA (approved):

k-doering-NOAA:
Alright, I think this PR is ready! Given there is an unresolved issue in the main branch currently being worked on, though, perhaps we should sort that out first before merging this in?

Rick-Methot-NOAA:
Kathryn,
Do you mean #152, #159 or which other issue is unresolved?

k-doering-NOAA:
I meant PR #160 (which addresses #154 ). It was merged in before the full jenkins run was conducted and I think caused those 2 model run failures and the 0 se values on a few forecasting quantities that we are now seeing on the main branch.

Rick-Methot-NOAA:
problem was this:
bigskate and yelloweye used F_std as exploitation rate
denominator for exploitation rate calculation is smrybiomass
as I messed with smrybio calculation it got buried in code that was for save_for_report==1
but F_std is needed for derived quantities, so was affected

PR #166, opened 2021-04-27
Natmort again

Rick-Methot-NOAA:
ready to see if re-arrangements needed for Maunder M (new natmort option=5) have broken anything else.

### DIFF
--- a/SS_benchfore.tpl
+++ b/SS_benchfore.tpl
@@ -1439,7 +1439,10 @@ FUNCTION void Get_Forecast()
           get_growth2(y);
         }
       if(timevary_MG(y,1)>0) get_natmort();
-      if(timevary_MG(y,3)>0) get_wtlen();
+      if(timevary_MG(y,3)>0){
+      	get_wtlen();
+        if(Hermaphro_Option!=0) get_Hermaphro();
+      } 
       if((timevary_MG(y,4)>0 || timevary_MG(endyr+1,4)>0) && Fcast_Loop_Control(3)!=3) 
       	{get_recr_distribution();}
       if(timevary_MG(y,5)>0) get_migration();
--- a/SS_benchfore.tpl
+++ b/SS_benchfore.tpl
@@ -1362,8 +1362,9 @@ FUNCTION void Get_Forecast()
         get_growth3(y,t,s, subseas);  //  in case needed for Lorenzen M
         Make_AgeLength_Key(s, subseas);
       }
-      if(s==spawn_seas) Make_Fecundity();
+//      if(s==spawn_seas) Make_Fecundity();
     }
+      get_mat_fec();
   }
 
     for (y=endyr+1;y<=YrMax;y++)
@@ -1506,7 +1507,8 @@ FUNCTION void Get_Forecast()
             }
             else if(timevary_MG(y,2)>0 || timevary_MG(y,3)>0 ||  bigsaver==1 )
             {
-               Make_Fecundity();
+//               Make_Fecundity();
+  get_mat_fec();  //  does just spawn season and subseason using ALK calculated just above
                for (g=1;g<=gmorph;g++)
                if(use_morph(g)>0)
                {
--- a/SS_biofxn.tpl
+++ b/SS_biofxn.tpl
@@ -1042,30 +1042,25 @@ FUNCTION void get_natmort()
             Maunder_Lmat = natMparms(3,gp);    //  constant for juvenile mort
             Maunder_Mmat = natMparms(4,gp);    //
             if(natM_5_opt<=2){	//use the SS mat50% and mat_slope parameters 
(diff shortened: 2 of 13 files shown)


Rick-Methot-NOAA:
Summary of code changes:
a.  SS_biofxn.tpl:  
- deprecate Make_Fecundity() and move its contents into new function get_mat_fec(); 
- split previous get_wt_len() into some code going into get_MGsetup(), some staying in get_wt_len(), and some moving into the new get_mat_fec().  This achieves a cleaner organization of the functions.  We should test fidelity of code for uncommonly used maturity-fecundity options
b.  SS_biofxn.tpl:
- introduce new natural mortality option (5) developed by Mark Maunder.  It links young fish M to fish size, and old fish M to maturity.  Has 3 sub-options.  Needs documentation which we should request from Mark.  write_ssnew labels this option as a BETA feature as performance testing has not been done.
b.  SS_prelim, SS_popdyn, SS_benchmark: 
-  re-arrange order of operations such that get_mat_fec is called before get_natmort.   This was tricky.  Needs testing for seasonal models, empirical bodywt models

Rick-Methot-NOAA:
@k-doering-NOAA :  The run-no-est failed and the output showed several se values at 0.0, even for simple.  On my local machine, the new code runs simple and matches the test value for the se of fore_catch_2011.  How can we dig deeper into the problem on github test run?

k-doering-NOAA:
@Rick-Methot-NOAA the no-est job runs with no hessian, which is why there are zeroes for sds. So sds are not causing the job to fail, it is the Widows model which has some likelihood values and Bratio_2014 that has changed. During the github action run, there is some text printed to the console regarding which comparison didn't pass, e.g., : https://github.com/nmfs-stock-synthesis/stock-synthesis/runs/2443528380?check_suite_focus=true#step:10:51

If you download the artifact from the failing job (should download a file called result_textfiles.zip) and unzip it, there should be test_failed.txt. This lists the comparison  values for only the models that didn't pass the checks. The other file, compare_test.txt, contains results for all the models that ran. 

Looking at github actions output can be pretty confusing at first, so feel free to let me know if you have more questions or if screensharing would be helpful.

Rick-Methot-NOAA:
Got it.  I missed the -nohess aspect of the comparison.

k-doering-NOAA:
@Rick-Methot-NOAA , so the run with no estimation runs from the par file - which means some of the math behind the model is changing. Isn't this a red flag even if the model is unstable?

Also, I think it would be good to run the full jenkins routine (runs including estimation and hessian)  on the natural mortality branch before merging, as it did catch some things last time we had an open PR. I will start it now and report back.

Rick-Methot-NOAA:
Yes, please do that full run before we merge.

The way to test widow is to:
a) get par from the base model
b) run base model with stopph 0 to get report
c) run testmodel with stopph 0 to get report
d) detailed compare of those reports

Can you do a,b,c and then I/we can do d?

k-doering-NOAA:
Thanks Rick, I did a, b, and c using the current main branch as the base model (which i confirmed has the same results as the reference run in the ss-test-models repository). I see the same differences that the no est run caught, which shows up as differences in additional derived values also. Here are the ss_summary files since they are smaller, but I can send the full report files if that would be helpful.
[ss_summary_files.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/6387558/ss_summary_files.zip)

Rick-Methot-NOAA:
Thanks.  I will need the full report.sso files so I can compare the
detailed age-structured tables.
I fear that I will not see anything different there, so the step after that
is to let models run to completion starting from the same place (make sure
jitter is off) and then do a detailed compare of parmtrace.sso to see when
during the run they diverge.

Rick

k-doering-NOAA:
I see, here are the report files.
[Report.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/6387690/Report.zip)

Rick-Methot-NOAA:
Just what I needed.  I got totally involved with a major writing assignment
this afternoon so set this aside for a bit.

The comparison narrowed the problem down immediately.  I see differences in
fecundity-at-age beginning in the first year, but body weight-at-age is
same.  SO the problem must be in my re-factoring of ss_biofxn where
fecundity-at-age is calculated.

I have 4 hr Science Board calls on Wed and Thurs, so might take a few days
to get this fixed.

Rick

k-doering-NOAA:
Ok, good to know!

Also, I checked on the full Jenkins run this morning, and it didn't pick up on any additional issues besides the one with Widow.

k-doering-NOAA:
@Rick-Methot-NOAA I see you pushed a new commit and looks like it passes the build-no-est job! Would you like me to rerun the full jenkins model runs or is there additional work you would like to do first?

PR #173, opened 2021-05-05
This fix addresses discrepancies in variance of SSB_unfished

Rick-Methot-NOAA:
tested with big_skate and with simple models.  Variance of SSB_unfished now matches variance of SSB_virgin as it should in these examples.  Ready for testing.

### DIFF
--- a/SS_ALK.tpl
+++ b/SS_ALK.tpl
@@ -16,7 +16,6 @@ FUNCTION void Make_AgeLength_Key(const int s, const int subseas)
    imatrix ALK_range_use(0,nages,1,2);
    if(ALK_subseas_update(ALK_idx)==1) //  so need to calculate
    {
-
    ALK_subseas_update(ALK_idx)=0;  //  reset to 0 to indicate update has been done
    gp=0;
     for (int sex=1;sex<=gender;sex++)
--- a/SS_benchfore.tpl
+++ b/SS_benchfore.tpl
@@ -1363,8 +1363,9 @@ FUNCTION void Get_Forecast()
         Make_AgeLength_Key(s, subseas);
       }
 //      if(s==spawn_seas) Make_Fecundity();
+      if(s==spawn_seas) get_mat_fec();
     }
-      get_mat_fec();
+//      get_mat_fec();
   }
 
     for (y=endyr+1;y<=YrMax;y++)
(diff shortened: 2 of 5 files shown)


Rick-Methot-NOAA:
Only failure is regarding the better result obtained for vermillion snapper, so actually a success.
However:  did testing stop at that failure, or were all models run?
Also, why did the previous testing only fail on big_skate?  It seems that at least simple should have failed also.  Is the tolerance not tight enough to detect the differences? @k-doering-NOAA

k-doering-NOAA:
Hi @Rick-Methot-NOAA, the model run and comparison is done for all models (of the fast running set) regardless of if one fails. So you are right that it is just Vermillion that is "failing", which is due to finding a better likelihood.

check out the all_changes.csv file within the artifact - it lists all changes from the refrence run, even ones below the tolerance that github actions will flag as "failed".

There is still the tiniest difference (ratio between values is 1.0000001) in SSB_unfished_se for Big skate, which seems negligable. Interestingly, there is no difference for Simple in these quantities.

k-doering-NOAA:
I missed one of your questions: looking at [this previous run](https://github.com/nmfs-stock-synthesis/stock-synthesis/runs/2504037489) which basically tests the code as it is in the current main branch, I don't see a difference for Simple in the SSB_unfished_se value in the all_changes.csv file. So I think this means that it is not showing up in the main branch?

Rick-Methot-NOAA:
It seems to me that "simple" should have failed when I first pulled the
natmort branch into main.

k-doering-NOAA:
I just checked the [jenkins run after merging natmort into main](https://jenkins.nwfsc2.noaa.gov/job/stock-synthesis-compare/141/artifact/run_R/compare_test.txt) and it looks like the results for Simple were exactly the same as the reference run, while Big Skate showed the SSB_unfished_se difference. I'm not sure why Simple shows the issue locally.

Rick-Methot-NOAA:
And I thought I had checked using simple thoroughly before doing that pull
request.  Note past me to have done one more code tweak after I had
finished all testing.
Anyway, I now see no reason not to complete this merge.  I will let you do
it when you are ready also.
Rick

k-doering-NOAA:
@Rick-Methot-NOAA , the only reason I see not to do the merge is that Jenkins is still running additional models (the runs that take more time) - unless it hinders your progress, I suggest waiting to merge this until it finishes running later today/overnight?

Does that seem ok? I think it is unlikely to fail given the models we are seeing issues with are all included in the new github actions run-with-est job, but I still think it might be helpful to continue Jenkins runs for a little bit, until we decide they are no longer adding value to our process.

Rick-Methot-NOAA:
Concur.  continue Jenkins run and merge after success there.

k-doering-NOAA:
Jenkins success, so merged in!

PR #178, opened 2021-05-12
remove_filelist_and_add_SS_label_file

Rick-Methot-NOAA:
This  branch:
a. removes some old code that was commented out;
b. removes the SS_filelist.txt file and the program_flow.tpl file
c.  adds and updates the info from SS_filelist.txt into the top of each file in markdown format and designated with the symbol: SS_Label_file.
There is no effect on the operational code.


k-doering-NOAA:
Nice work @Rick-Methot-NOAA! I think for things like this, it makes sense just to let the checks finish just in case an accidental change to the code snuck in, then if they all pass, I think you can go ahead and merge in your own work (without anyone else reviewing it). Does that sound like a good process to you? Just wanted to clarify since I think this is the first PR that you have submitted that doesn't have code changes.

Rick-Methot-NOAA:
Thanks for checking.  That was my intention.

Issue #186, opened 2021-06-18
trap for duplicate index observations


iantaylor-NOAA:
As discussed in https://github.com/iantaylor-NOAA/Lingcod_2021/issues/59#issuecomment-864229746, the SS model files in [lingcod_model_with_error.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/6678960/lingcod_model_with_error.zip)
 from the still-in-process-with-lots-left-to-do lingcod assessment produced the following error.
```
Initial statistics: 1 variables; iteration 0; function evaluation 0; phase 1
Function value   1.3995186e+05; maximum gradient component mag  -3.9184e+04
Var   Value    Gradient   |Var   Value    Gradient   |Var   Value    Gradient
  1 -0.05710  -3.9184e+04 |
1 2 -log(L): 141340  Spbio: 63999.6 47363.9
1 3 -log(L): 143445  Spbio: 63999.4 47363.7
1 4 -log(L): 146155  Spbio: 63999.4 47363.7

Function minimizer: Step size  too small -- ialph=11 5 -log(L): 149373  Spbio: 63999.6 47363.9
1 5 -log(L): 149373  between
2 6 -log(L): 153012  Spbio: 63999.6 47363.9
```
Dave Fournier said (on [ADMB Users email list in 2010](http://lists.admb-project.org/pipermail/users/2010-December/000860.html))
> It means that the function minimizer can not find a step size which decreases the function.

Indeed the -log(L) just rises steadily.

I tracked it down to duplicate index observations added by accident. The fleet 4 index with comments like "#_index_recORCPFV" (starting on line 584 of ling_data.ss) should have had negative year associated with them.

I don't know if this is a general property of any model with duplicate index observations for the same fleet/year, but stopping or warning when duplicates occur would help track down the error.

Rick-Methot-NOAA:
this code in readdata should have produced a warning:

        if(data_time(ALK_time,f,1)<0.0)  //  so first occurrence of data at ALK_time,f
          {data_time(ALK_time,f)(1,3)=timing_r_result(1,3);}  // real_month,fraction of season, year.fraction
        else if (timing_r_result(1) !=  data_time(ALK_time,f,1))
[...]

PR #187, opened 2021-06-28
issue #186; plus remove leading spaces

Rick-Methot-NOAA:
@iantaylor-NOAA 
Ian, this should be completely straightforward, but please take a look from a r4ss perspective.


iantaylor-NOAA:
Good news. There were no issues at all with loading a Report.sso created by the ss.exe from the artifact from the GitHub Action associated with this pull request.  The read.table() function in R which is used to initially read the output files appears to be robust to extra leading whitespace. Go ahead and merge when you're ready.

PR #196, opened 2021-08-19
fix lorenzen M for seasons

Rick-Methot-NOAA:
change should only affect Lorenzen M models that use seasons


k-doering-NOAA:
We don't have a seasonal Lorenzen M model in our test set, hence all checks passing! @Rick-Methot-NOAA, let me know if you would like me to do one off tests

Rick-Methot-NOAA:
I did unit test of this feature and satisfied that it is working as
advertised.  I will approve the pull request.

PR #199, opened 2021-08-25
finish fix for combined age-length selex

Rick-Methot-NOAA:
Kelli,
Please check that the se for combined age-length selectivity seems reasonable.  I tried the simple example after increasing N platoons to 5 (just to test the indexing) and I get this:
AgeLenSelex_std_1_GP1_Fem_A_0 0.00529027 0.000715725
AgeLenSelex_std_1_GP1_Fem_A_1 0.0171669 0.00210697
AgeLenSelex_std_1_GP1_Fem_A_2 0.0503977 0.00558243
AgeLenSelex_std_1_GP1_Fem_A_3 0.127717 0.0127583
AgeLenSelex_std_1_GP1_Fem_A_4 0.260027 0.0234945
AgeLenSelex_std_1_GP1_Fem_A_5 0.410182 0.0337065
AgeLenSelex_std_1_GP1_Fem_A_6 0.526735 0.0396442
AgeLenSelex_std_1_GP1_Fem_A_7 0.59982 0.0416719
AgeLenSelex_std_1_GP1_Fem_A_8 0.644169 0.0416408
AgeLenSelex_std_1_GP1_Fem_A_9 0.673059 0.0407983
AgeLenSelex_std_1_GP1_Fem_A_10 0.693696 0.0397205
AgeLenSelex_std_1_GP1_Fem_A_11 0.709521 0.0386378
Thanks,
Rick

### DIFF
--- a/SS_objfunc.tpl
+++ b/SS_objfunc.tpl
@@ -1045,6 +1045,7 @@ FUNCTION void Process_STDquant()
 
 //  SS_Label_7.8  get extra std quantities
     // selectivity
+    //  f = Do_Selex_Std
     if(Selex_Std_Cnt>0)
     {
       for (i=1;i<=Selex_Std_Cnt;i++)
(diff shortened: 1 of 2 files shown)


kellijohnson-NOAA:
If an age in the age-vector for `Vector of bin indexes` is beyond the max age, then the max age is used. There was no warning in warnings.sso. This is fine, but I wanted to make sure it is intentional. I also put in 2 ages older than the max age and it led to the max age information being printed two times.

Rick-Methot-NOAA:
That behavior is intentional and glad it worked.  A warning would be good
to add.

kellijohnson-NOAA:
@Rick-Methot-NOAA I checked that the estimated uncertainty went up and down in the correct directions when sample sizes for lengths or ages decreased or increased. Intuitively that is all that I can think of to do.

The documentation for what the output means in terms of options 1, 2, and 3 is not well documented; though I think that options 1 and 2 are intuitive. For option 3, is there a description of what a value for a single age means in terms of age-length selectivity?

Let me know if you would like more testing here. Such as, should I try different selectivity forms or is the small amount that I did sufficient?

Rick-Methot-NOAA:
Thanks Kelli.  That seems sufficient.
It terms of documentation, I think we need to refer to the selectivity section and not to do much more here in the extra variance section.  The quantity itself is the selectivity-at-age that results from Sel(age) * dot product(sel(length),distribution(length@age)).  That should be described in the selectivity section.

kellijohnson-NOAA:
Yes, I agree the selectivity section is the best section, that is where I went to search for information. I also think it would be good to specify that all four values are still needed even if the first value is 0 which turns off the sd calculation for selectivity and you have to remove the line of ages or lengths for reporting.

Rick-Methot-NOAA:
created warning per Kelli's suggestion.
Proceeding with squash and merge

Issue #202, opened 2021-08-27
Change version info in github action


k-doering-NOAA:
Do this to make it clear that the github action job is a tester version.

Note we will probably want a separate job for exes that are formally released (releases and distributed fixes)

k-doering-NOAA:
Thanks @kellijohnson-NOAA and @Rick-Methot-NOAA ! These are all great thoughts. I really like the idea of making more things automated.

I was thinking maybe everything would just be labeled experimental (maybe with the last commit number), unless it is tagged and in the main branch? That way, when we are ready to do a release, we can just add the tag manually, as we normally do (if it is a prerelease, maybe we put that in the tag?) and the github action will take care of the rest? This isn't quite as automated as what  Kelli is proposing, but I think it would basically allow us to work in the same way while automating this extra piece (the version numbering).

k-doering-NOAA:
> This would create a lot more MINOR and PATCH increments in between releases and it begs question of whether what we label as a "release" is automatically a MAJOR increment. So far, MAJOR is stuck at .30 and I do sometimes think about what would constitute .31. Conceptually, 4.0 will be FIMS, but I'll refrain from calling FIMS SS4.00.00.00

Good point...I wonder if bumping the major version would indicate to folks a big I/O change, which we (thankfully) haven't had? So perhaps given the way we have been doing versioning, it makes sense to be stuck at .30?

k-doering-NOAA:
@Rick-Methot-NOAA , how should we implement this? Here are two options I am thinking about, but feel free to suggest another

1. by default, says experimental
In the files, 
```
#V3.30.experimental; not an official version of SS;_safe;...
```
2. by default, says a placeholder, like 3.30.xx.yy.
In the files, 
```
#V3.30.xx.yy;_safe;...
```

I think this matters because this is what would show up when compiling locally. I'm leaning towards 1, but what are your thoughts?

Rick-Methot-NOAA:
I already am doing something like (1), but using beta rather than
experimental.

3.30.17.beta

but should it be beta developments after .17

or 3.30.18beta   as a beta for .18 in development?

k-doering-NOAA:
Hmm, good point! I think leaving it as 3.30.17.beta makes sense given how we number fixes currently.

[...]

PR #209, opened 2021-09-13
build, issue #202: update gh actions build to automate version numbering

k-doering-NOAA:
This PR addresses #202. Now, version info is automatically added based on tags. If a tag with the same commit has been added and has the right format (v3.30.xx.yy, v3.30.xx, or v3.30.xx-prerel), the version as in the tag will be added to the SS version info file before compilation. Otherwise, the last xx version will be used, with .yy = .beta (e.g., a commit after v3.30.17 but before v.3.30.18 and that isn't tagged would be labeled v.3.30.xx.beta).

Note this change does not affect the date of compilation, which will be the true date of compilation due to the use of __DATE__ in the version info.

### DIFF
--- a/.github/workflows/build-centos.yml
+++ b/.github/workflows/build-centos.yml
@@ -1,8 +1,12 @@
 name: build-centos
 
-# Controls when the action will run. Triggers the workflow on push or pull request
-# events but only for the main branch
-on: [workflow_dispatch]
+# on: Controls when the action will run. workflow_dispatch is a manual trigger.
+# This workflow also runs when a tag begining with v is pushed.
+on: 
+  workflow_dispatch:
+  push:
+    tags:
+      - "v*"
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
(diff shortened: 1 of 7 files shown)


Rick-Methot-NOAA:
the proposed change to version_info means that any local compilation of SS will now show version number as 3.30.xx.yy because the local make file does not have the version incrementing capability.  So we will be highly dependent on using this capability well.  That's a good thing as it prevents me from distributing clandestine versions.
Also, the proposal depends upon correctly formatted tag being read.
Please provide a few examples.

k-doering-NOAA:
Tags should look like the following (note I can add this to the developer's wiki page as well when/if we merge it in) :
- For a prerelease: v3.30.xx-prerel (e.g., v3.30.18-prerel)
- For a release: v3.30.xx (e.g., v3.30.18)
- For a fix: v3.30.xx.yy (e.g., v3.30.18.01)

Other tagging formats will cause the exe to be marked as v3.30.unknown.unknown

Rick-Methot-NOAA:
Ahh.  So the tag the user creates IS the version.  So the action does not read the old version and increment the number, you simply enter the version info as the tag.
so entering tag as "v3.30.18.release"
would create version_info as 3.30.18.release.

OR, am I still not getting what is user input and what is automatically incremented?

Rick-Methot-NOAA (approved):
I am satisfied that these changes will improve the workflow for the project.  I approve.

PR #217, opened 2021-10-15
Updated file comments

nschindler-noaa:
This makes a few changes to comments so that, when using getlables.exe, the result is markdown text.
Tested ss.tpl after copying all the files together.


Rick-Methot-NOAA:
Thanks Neal

PR #218, opened 2021-10-22
new_filename_and_folder_for_outputs

Rick-Methot-NOAA:
will write to /ssnew and /sso folders if they exist and should write separate report and compreport files in mceval (not tested).

### DIFF
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -34,7 +34,6 @@ FUNCTION void write_bigoutput()
         anystring2+=anystring+".sso";
         SS2out.open(anystring2);
         report_sso_filename=anystring2; //  save so can be reopened in append mode
-
         anystring2=sso_pathname+"CompReport_mce_";
         if(mceval_counter<10)
         {anystring2+="000";}
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -20,9 +20,24 @@ FUNCTION void write_bigoutput()
       else
       {
         sprintf(anystring, "%d", mceval_counter);
-        anystring2=sso_pathname+"Report_mce_"+anystring+".sso";
+        if(mceval_counter<10)
+        {anystring2=sso_pathname+"Report_mce_000"+anystring+".sso";}
+        else if(mceval_counter<100)
+        {anystring2=sso_pathname+"Report_mce_00"+anystring+".sso";}
+        else if(mceval_counter<1000)
+        {anystring2=sso_pathname+"Report_mce_0"+anystring+".sso";}
+        else
+        {anystring2=sso_pathname+"Report_mce_"+anystring+".sso";}
         SS2out.open(anystring2);
-        anystring2=sso_pathname+"CompReport_mce_"+anystring+".sso";
+
+        if(mceval_counter<10)
+        {anystring2=sso_pathname+"CompReport_mce_000"+anystring+".sso";}
+        else if(mceval_counter<100)
+        {anystring2=sso_pathname+"CompReport_mce_00"+anystring+".sso";}
+        else if(mceval_counter<1000)
+        {anystring2=sso_pathname+"CompReport_mce_0"+anystring+".sso";}
+        else
+        {anystring2=sso_pathname+"CompReport_mce_"+anystring+".sso";}
         SS_compout.open(anystring2);
       }
 
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -1078,7 +1078,7 @@ FUNCTION void write_bigoutput()
     else
     {
     for (gp=1;gp<=N_GP;gp++) {SS2out<<" _ ";}
-
+    if(Hermaphro_Option!=0) {for (gp=1;gp<=N_GP;gp++) {SS2out<<" _ ";}}
     }
     SS2out<<" "<<Bio_Comp<<" "<<Num_Comp;
     if(s==1 && y<=endyr) {Smry_Table(y,15)+=smryage;}  // already calculated for the forecast years
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -8,8 +8,33 @@
  /*  SS_Label_FUNCTION 40 write_bigoutput */
 FUNCTION void write_bigoutput()
   {
-  SS2out.open ("Report.sso");   // this file was created in globals so accessible to the report_parm function
-  ofstream SS_compout("CompReport.sso");
+
+    struct stat pathinfo;
+    adstring pathname="";
+    if( stat( "./sso", &pathinfo ) != 0 )
+    {
+      pathname="";
+    }
+    else
+      {
+        pathname="./sso/";
+      }
+
+      if(mceval_counter==0)
+      {
+        anystring=pathname+"report.sso";
+        SS2out.open(anystring);   // this file was created in globals so accessible to the report_parm function
+        anystring=pathname+"CompReport.sso";
+        SS_compout.open(anystring);
+      }
+      else
+      {
+        sprintf(anystring2, "%d", mceval_counter);
+        anystring=pathname+"report_mce_"+anystring2+".sso";
+        SS2out.open(anystring);
+        anystring=pathname+"CompReport_mce_"+anystring2+".sso";
+        SS_compout.open(anystring);
+      }
 
   SS2out<<version_info(1)<<version_info(2)<<version_info(3)<<endl<<version_info2<<endl;
   time(&finish);
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -9,30 +9,19 @@
 FUNCTION void write_bigoutput()
   {
 
-    struct stat pathinfo;
-    adstring pathname="";
-    if( stat( "./sso", &pathinfo ) != 0 )
-    {
-      pathname="";
-    }
-    else
-      {
-        pathname="./sso/";
-      }
-
       if(mceval_counter==0)
       {
-        anystring=pathname+"report.sso";
+        anystring=sso_pathname+"Report.sso";
         SS2out.open(anystring);   // this file was created in globals so accessible to the report_parm function
-        anystring=pathname+"CompReport.sso";
+        anystring=sso_pathname+"CompReport.sso";
         SS_compout.open(anystring);
       }
       else
       {
         sprintf(anystring2, "%d", mceval_counter);
-        anystring=pathname+"report_mce_"+anystring2+".sso";
+        anystring=sso_pathname+"Report_mce_"+anystring2+".sso";
         SS2out.open(anystring);
-        anystring=pathname+"CompReport_mce_"+anystring2+".sso";
+        anystring=sso_pathname+"CompReport_mce_"+anystring2+".sso";
         SS_compout.open(anystring);
       }
 
@@ -643,7 +632,6 @@ FUNCTION void write_bigoutput()
     }
     SS2out<<endl;
   }
-  ofstream post_vecs("posterior_vectors.sso",ios::app);
   post_vecs<<runnumber<<" 0 "<<obj_fun<<" F/Fmsy_stdev ";
   for (j=1;j<=N_STD_Yr_F;j++)
   {
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -11,6 +11,7 @@ FUNCTION void write_bigoutput()
 
       if(mceval_counter==0)
       {
+        cout<<" writing big output now "<<endl;
         anystring=sso_pathname+"Report.sso";
         SS2out.open(anystring);   // this file was created in globals so accessible to the report_parm function
         anystring=sso_pathname+"CompReport.sso";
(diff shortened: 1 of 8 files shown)


Rick-Methot-NOAA:
Kelli, Chantel and Ian - could I get you opinion on what to do with the data.ss_new while I work to get report for each mceval working.
options are:
a.  status quo (Rick hates it)
b.  status quo but to a sub-folder if subfolder exists
c.  separate files for data_echo.ss_new, data_expval.ss_new, the all bootstraps appended in 3rd file
d.  all in separate files

Right now the branch does (b) and (d)

@kellijohnson-NOAA @chantelwetzel-noaa @iantaylor-NOAA

kellijohnson-NOAA:
(a) - lots of workflows depend on it
(b) - I think the overhead of changing everyone's simulation code is not worth it for this option. What is the difference if the file is in the main directory versus a subdirectory, the structure is still the same? Are you trying to get at the point that the output file is not the same as the input file where the other ss_new files are?
(c) - I would change the extension of anything that is not the echoed data to something different than ss_new. So my proposal here would be data.ss_new, data_expected.ss, data_boot.ss where the .ss could be changed to something different for the latter two files.
(d) - has anyone checked the overhead of writing 100 bootstrap files versus 1 file with 100 data sets?

iantaylor-NOAA:
I think (d) would be my ideal, but in the simulations I've worked on (long ago), I wanted to the expected recdev time series for each bootstrap so never wrote more than 1 per model run in which case (c) and (d) are essentially the same. 
For (c) and (d), I like the filenames proposed by @kellijohnson-NOAA in the comment above https://github.com/nmfs-stock-synthesis/stock-synthesis/pull/218#issuecomment-965414356. 

I know that past workflows have all used the status-quo approach but don't have a good sense of how many of those are likely to ever be repeated. Adapting to any of these options also shouldn't be very hard, so I'm not sure that backward compatibility is important if it adds complexity to the code.

As for run time, I would guess that the bootstrap and file-writing process is much faster than the model estimation step that is surely part of almost all bootstrap analyses so it's not worth trying to hard to optimize the run time.

Rick-Methot-NOAA:
Lets clarify what we mean by bootstrap.  There is the bootstrap process
inside of SS3 that is post-estimation and the only slow part of it is doing
the multinomial sampling for all the comps.  We are talking about the
output from that process.  Then there is the ss3sim process that repeatedly
calls the entire ss3 process, that takes longer and each of the ss3sim
calls creates the entire set of ss3 outputs.

chantelwetzel-noaa:
I generally agree with Ian and Kelli's suggestions.  I would prefer option
c or d with the naming structure proposed by Kelli for clarity.  Similar to
Ian, I have only done simulations where I generate a single bootstrap at a
time, so don't have much of an opinion on generating all bootstraps in a
single file or generating separate files.

Chantel Wetzel, PhD

*Pronouns: she/her *

*Fishery Resource Analysis and Monitoring DivisionNorthwest Fisheries
Science Center*
*National Marine Fisheries Service*
*National Oceanic and Atmospheric Administration*

k-doering-NOAA:
We will let this one sit for a bit, as we don't want to change the subfolder and file structure in the main branch yet. However, we plan to merge in before the 3.30.19 release.

Rick-Methot-NOAA:
Kelli,    @kellijohnson-NOAA 
this branch now produces a separate report and compreport file for each mceval.  Files have names like:
Report_mce_25.sso for the 25th mceval iteration. These reports will be found in the runfile directory unless the user creates a ./sso subdirectory. Please take a look and let me know if this meets the needs of the hake assessment. The branch also implements the previously discussed changes to data.ss_new.  These go to subfolder ./ssnew if it exists and have names of:
data_echo.ss_new, data_expval.ss, and data_boot_1.ss through data_boot_N.ss.
Rick

kellijohnson-NOAA:
I am testing this out today.

kellijohnson-NOAA:
@Rick-Methot-NOAA is the functionality intentional to have posteriors.sso be a blank file? Also, as written the system works when I run it with standard MCMC but @cgrandin was having trouble when using @Cole-Monnahan-NOAA's adnuts. Chris is looking into this more as will I.

Rick-Methot-NOAA:
Nothing intentional was done to change posteriors, so I will check on that.

No idea how adnuts would be affected.

Rick-Methot-NOAA:
@kellijohnson-NOAA  code is now writing correctly to posteriors.sso.  I do not know how to investigate the adnuts issue.

kellijohnson-NOAA:
Great. I will distribute the new executable to the team.

I checked with Chris and he said that it ended up being fine with adnuts. I think he panicked because the files are not written until the end. Sorry for the false alarm there.

Rick-Methot-NOAA:
@k-doering-NOAA  Please merge this into main after you are able to adjust the workflow to deal with the new filenames.

k-doering-NOAA:
Ok, sounds good, Rick. I'm working on this now; hope to finish by end of the day monday at the latest!

cgrandin:
Hi all,

Thanks for getting this done and so quickly!
One little thing though is that the `posteriors.sso`, `deriver_posteriors.sso`, etc files are located in the `sso` folder. This may break some other folks code if they try to run with an `sso` folder and have current code that reads those files from the parent `mcmc` folder. Would it be possible to move those files back to the parent directory of `sso`?

Thanks

Rick-Methot-NOAA:
Hi Chris.
That would be possible, but it seems overly complicated to have some sso
files in the main folder and others in the sso folder. Do you have a
complete list of which you would want where?
All sso files go to the main folder if the sso folder has not been created
by the user prior to the ss3 run starting.
Rick

k-doering-NOAA:
Some observations on this PR:

- Tests are fixed in the main branch, so when this gets merged in, the checks should work again. I tested by making a new branch from this one, then pulling in the main branch and running the tests. They were passing, as indicated by the blue arrow of the last commit (https://github.com/nmfs-stock-synthesis/stock-synthesis/tree/output_to_separate_files_check)
- I tested the code locally and manually again to see what happens under various folder scenarios. I learned that if a subfolder called sso exists, then sso files are written there; if a folder called ssnew exists, then the .ss_new filew are written there.
- The new data files are called data_echo.ss_new, data_expval.ss, and dat_boot_x.ss . I know that Kelli liked the expval and boot not to have ss_new extensions, but I wonder if it is then confusing that the filse do not have the .ss_newextension but live in the folder labeled ssnew anyway? I personally think it is less confusing to have the files with the ss_new extension since they are (sometimes) written to the ssnew folder.
- A picky thing about bootstrap file names: the first is called dat_boot_1.ss ; the 10th would be dat_boot_10.ss; we could potentially add leading zeros instead to help order them (e.g., dat_boot_001.ss; dat_boot_010.ss). This would assume most people don't have more than 999 bootstrap files. I'm ok with either, just wanted you to think about which names you prefer, Rick.
- @kellijohnson-NOAA , when you have the chance, could you list the steps on how to test this out with mcmc (for people who don't really use mcmc in admb, like me? :) ) I tried running, but wasn't successful at getting the files to show up as they should. I'm guessing this is user error on my part

cgrandin:
Hi @Rick-Methot-NOAA ,

I've broken down my logic here, maybe I am missing something. If so, my apologies!

If I run this:
`ss -mcmc 8000 -nuts`, and there is no `sso` directory, the behaviour is that the `posteriors.sso`, `derived_posteriors.ss`, etc files are created in the directory in which the mcmc was called. 

If I run this (same thing):
`ss -mcmc 8000 -nuts`, and I have an `sso` directory, the `posteriors.sso`, `derived_posteriors.ss`, etc files are not created in the directory in which the mcmc was called. Instead, they are located in the `sso` directory.

My question is, if I'm correct, is this how you really want it, considering they will be in two different places depending on how SS was run?

One other thing is that if the user is using Windows File explorer (not fluent with the command line), they have to scroll through thousands of files to find the files. They could filter the results but they would have to do that for each file one by one to see that they exist and have some size after the run. It's not an issue for me but I can see that it would be for some.

For Hake we are good like it is, I already our code working with this, I'm just trying to help other users.
Thanks Rick

Rick-Methot-NOAA:
Hi Chris,
Now I am confused as this option was created in order to help organize the many new files requested by the hake team.
There is no change to posteriors.sso, etc. so I do not see why that is an issue, especially the user can choose a workflow in which they create a sso folder and hence know to look there for output, or not create the folder and look for the output in the same place they always looked for it.
The specific request from the hake team was for report.sso and compreport.sso from each mceval call.  Appending all of these into one file each seemed very unwieldy, very large, and very difficult to see if a particular run existed. Hence I have put each into separate named files, with the name representing the mceval index number.  Having this output is new, so some new workflow was needed anyway for you to use them.
What do others think about this approach?

cgrandin:
I think it's totally fine, we have everything working the way it is.
The way the file structure is laid out is fine for our team.

Thanks!

k-doering-NOAA:
@Rick-Methot-NOAA I just noticed that the option 2 for MCMC detail is not yet described in starter.ss_new comments:

```
2 # MCMC output detail: integer part (0=default; 1=adds obj func components); and decimal part (added to SR_LN(R0) on first call to mcmc)
```

Rick-Methot-NOAA:
I will fix that.

Rick-Methot-NOAA:
In my final push on this issue I revised the numbered file names to have leading digits.
example:
data_boot_005.ss
Report_mce_0025.sso

SS_write_report.tpl
```diff
@@ -3298,6 +3326,8 @@ FUNCTION void write_bigoutput()
   SS2out<<" end selex output "<<endl;
   }  // end do report detail
   wrote_bigreport++;
+  SS2out.close();
```
kellijohnson-NOAA, line 3329: @Rick-Methot-NOAA For some reason this line leads to Dynamic Bzero NOT being written in Report.sso. At the top of the report file it says `Y Dynamic_Bzero report:59` but the section is not written.

Rick-Methot-NOAA: it should be in report.sso.
I intentionally turned off the profiling outputs for the MCEVAL outputs but did not intentionally change dynamic Bzero.  I will chenck on this also

kellijohnson-NOAA:
Thanks Rick, I checked and was able to successfully get 1002 samples ✔️.

One last mystery that is only happening in MCMC mode with this executable is in `AGE_SELEX report:32` the final year for `sel*wt` is wrong, it ends up being 2020 instead of 2021. If I run the executable in MLE mode everything is fine.

```
sel*wt 1 2020 1 1 1 2020_1_sel*wt 0 0.000112571 0.0109621 0.0703895 0.5076 0.280687 0.530073 0.549706 0.558731 0.59539 0.601531 0.651961 0.589622 0.780642 0.813207 0.869964 0.869964 0.869964 0.869964 0.869964 0.869964
```

Rick-Methot-NOAA:
are you seeing that discrepancy in report.sso, or in one of the mce reports?
The order of operations is getting quite complex, hence the problem with dynamic Bzero.

kellijohnson-NOAA:
@Rick-Methot-NOAA the problem with sel*wt... happens in both Report.sso and Report_mce_xxxx.sso if running in with -mcmc. If I just run `.\ss.exe` then those rows have the correct year.

Rick-Methot-NOAA:
Kelli,
Alas, I'm not going to solve the missing dynamic Bzero (and other things) today.  Some change I made in order to produce report.sso in the procedure section for mceval has messed with the code logic in a way that has disabled things (like dynamic Bzero) that get appended to the report.  So, your  holiday gift will be late.

Rick-Methot-NOAA:
Found the solution. Now I can relax

kellijohnson-NOAA:
Thanks Rick. I will look at this tomorrow. Merry Christmas 🎄 🎁 🎄

k-doering-NOAA:
Glad to see all these new fixes! Congrats. I am rerunning the checks on this branch : https://github.com/nmfs-stock-synthesis/stock-synthesis/tree/out_to_sep_check, which merged into the main branch so the newer gh action files could be used (which is the reason for the failing tests on this branch).

kellijohnson-NOAA:
Unfortunately, I am not able to get the latest fix to work with the hake model that I am testing it on. If I run in MLE I get dynamic bzero but there are still no values even though I have it turned on when running in MCMC. As well as the final year for the `sel*` output is final year - 1 not final year. Rick, is there something special that I should be doing in starter.ss? I tried turning on all output and turning on minimal output but calling dynamic bzero specifically and neither worked.

k-doering-NOAA:
The simple lorenzen model has a change in results when run with estimation, I think because the se for the forecast catch value decreased see [these model results](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/7780961/test_failed.csv)
). 

Rick, I think you remember you fixing something to do with forecast SE not too long ago, so I wonder if that is the reason? Perhaps the change is expected?

Rick-Methot-NOAA:
Kelli - you need only make this change in starter:

2 # MCMC output detail: integer part (0=default; 1=adds obj func components; 2= +write_report_for_each_mceval); and decimal part (added to SR_LN(R0) on first call to mcmc)

PR #223, opened 2021-11-09
Change abbreviation for Stock Synthesis

k-doering-NOAA:
Spells out stock synthesis or uses SS3 instead.

Also changes instances of STARTER.SS in the source code to starter.ss


k-doering-NOAA:
Merging this in, as there are now no conflicts and there are no functional changes to the code (just to docs/using ss3 in warning messages)

PR #224, opened 2021-11-12
use reusable workflows

k-doering-NOAA:
This PR changes to using reusable workflows located at https://github.com/nmfs-stock-synthesis/workflows/tree/main/.github/workflows , which allows having a centralized location for stock synthesis related workflows.

This PR is mostly for awareness; note that the exes are now built in a single job, call-build-ss3.yml, which will build the windows, linux, and mac exes. Note the linux exes are built on ubuntu instead of centos (the centos job was left just in case it is needed). The artifacts from these jobs are stored in a single artifact folder, which makes finding the exes more convenient.


Rick-Methot-NOAA:
I suggest that "get ADMB" should be a separate workflow. It is needed infrequently and when a new version of ADMB gets created then often is some change to the workflow involved.

A markdown doc with an overview of the various workflows would be helpful.  I remain a bit perplexed why pull requests generate six actions.

k-doering-NOAA:
The step "get admb" is simply installing the precompiled admb on the virtual machine. We need to do this every time Stock Synthesis is installed, so it actually is used frequently.

Good point about the documentation! - I need to update the wiki with this information.

Rick-Methot-NOAA:
OK.  I approve this pull request.

kellijohnson-NOAA:
There is an ability to "[cache](https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows)" part of a GitHub action and if getting / installing ADMB is using cloud minutes that could be used somewhere else we might think about looking into storing ADMB. I am not certain though if the storage of a cached item is largely for just using it quickly in between workflows if you have multiple workflow files you run every time or if it is for say days between the calling of a workflow.

k-doering-NOAA:
Thanks @Rick-Methot-NOAA  and @kellijohnson-NOAA !

Kelli, I forgot about the ability to cache, thanks for the reminder! In this case, ADMB takes 17 sec or less to install, so I don't think it is worth it. There are probably some other places, though where I should use caching...

Issue #205, opened 2021-09-01
possible bug in forecast calcs


Rick-Methot-NOAA:
in forecast, this code is within the season loop, so the initialize to 0.0 seems wrong.  This would only affect a multi-season model, so should be investigated with such:
if((Fcast_Loop1==2 || Fcast_Loop_Control(1)==1) && ABC_Loop==1) // get variance in OFL
{
Mgmt_quant(Fcast_catch_start+N_Fcast_Yrs+y-endyr)=0.;
for (f=1;f<=Nfleet;f++)
{
if(fleet_type(f)==1)
{Mgmt_quant(Fcast_catch_start+N_Fcast_Yrs+y-endyr)+=catch_fleet(t,f,2);}
else if (fleet_type(f)==3)
{} // survey
else if(bycatch_setup(f,2)==1) // bycatch
{Mgmt_quant(Fcast_catch_start+N_Fcast_Yrs+y-endyr)+=catch_fleet(t,f,2);}
}
}

PR #236, opened 2021-12-11
one line fix for issue #205

Rick-Methot-NOAA:
This problem affected forecast of OFL catch in a model with seasons
OFL catch contained only catch from Nseas
Fix involved deleting one line that was re-initializing that quantity


k-doering-NOAA:
Looks good to merge in, Rick, if you think so! Great it was just a one line fix.

PR #237, opened 2021-12-13
Github actions: only build tests on changes to tpl files

k-doering-NOAA:
Following https://github.com/nmfs-stock-synthesis/workflows/issues/15 , only build the stock synthesis on changes to the .tpl files. However, the option of manually triggering jobs remains, in case situations come up where builds are desired due to changes in other files.


k-doering-NOAA:
merging this in, as it is just a simple change of build triggers for github actions

PR #244, opened 2021-12-22
implement revision of F_Method 4 to allow constant F

Rick-Methot-NOAA:
simple change


k-doering-NOAA:
Thanks @Rick-Methot-NOAA! I had a look and think this can be merged in when you are ready.

The failing run-no-est tests can be ignored (they are due to assuming data_echo.ss_new is the name of the .ss_new file already); we know that no model results have changed because run-with-est has passed.

PR #249, opened 2022-01-06
rm extra empty line from control file

k-doering-NOAA:
causing issues with r4ss

### DIFF
--- a/SS_write.tpl
+++ b/SS_write.tpl
@@ -900,7 +900,9 @@ FUNCTION void write_Bzero_output()
       }
     }
 
-    SS2out<<endl<<"Report_Z_by_area_morph_platoon"<<endl;
+    SS2out<<endl<<"Report_Z_by_area_morph_platoon";
+    if(fishery_on_off==0) {SS2out<<"_1 No_fishery_for_Z=M_and_dynamic_Bzero";} else {SS2out<<"_2 With_fishery";}
+    SS2out<<endl;
     SS2out<<"Area Bio_Pattern Sex BirthSeas Settlement Platoon Morph Yr Seas Time Beg/Mid Era"<<age_vector <<endl;
     for (p=1;p<=pop;p++)
     for (g=1;g<=gmorph;g++)
--- a/SS_write.tpl
+++ b/SS_write.tpl
@@ -901,7 +901,7 @@ FUNCTION void write_Bzero_output()
     }
 
     SS2out<<endl<<"Report_Z_by_area_morph_platoon"<<endl;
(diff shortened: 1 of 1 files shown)


k-doering-NOAA:
@iantaylor-NOAA , if you have changes to add, feel free to push them here! All I did was (hopefully) get rid of the empty line between the table header and the column names for the Report_Z_by_area_morph_platoon report file table

Still to resolve is 
- [ ] Figure out why this table is appearing twice
- [ ] See if other changes are needed due to getting rid of a subheader line that was present in previous versions of SS3.

iantaylor-NOAA:
@k-doering-NOAA, your fix looks good to me. Seems like you could merge this change first to confirm that the r4ss github action gets working while we investigate the other issues.

iantaylor-NOAA:
I now see that the `test-r4ss-with-ss3` action will be conducted on this PR even if you leave it open, but merging would allow future branches from main (or existing branches that get rebased onto main) to avoid the error.

k-doering-NOAA:
Good point, @iantaylor-NOAA ; I think you are right that this simple change can be merged in once the checks run and we could work on the other issues in a separate branch. I'll merge in tomorrow morning, but feel free to merge it in tonight if you see it finishes

iantaylor-NOAA:
The `call-test-r4ss-with-ss3` workflow seems to be working now thanks to the blank line fix, thanks @k-doering-NOAA.

I think that the additional commit 96b1403 resolves the other two issues that @k-doering-NOAA identified.

The subheader separated the "With_fishery" and "No_fishery" tables which previous both fell under the main "Report_Z_by_area_morph_platoon" header. However, those two tables are now reported in separate places (similar to the other output in this section), but with the same header creating the appearance of duplicate tables.

Commit 96b1403 expands the headers to read
"Report_Z_by_area_morph_platoon_2 With_fishery"
and
"Report_Z_by_area_morph_platoon_1 No_fishery_for_Z=M_and_dynamic_Bzero"

A minor change to `r4ss::SS_output()` was already needed to deal with the different order of the tables so can now use these revised headers to distinguish the two parts.

Rick-Methot-NOAA:
thanks for cleaning up the mess I made when doing what I thought was a logical re-org of the tables.  I did it partly to elimiate an extra run through the timeseries.
Which reminds me; I think it useful to create another internal flag so we can run through the timeseries simulation without doing the call to get_exp_val.  This could speed up simulation work in general when data generation and logL is not needed

k-doering-NOAA:
I see, this all makes sense now. I ran a model locally just to make sure it works as we think it does, and looks great!

@Rick-Methot-NOAA  or @iantaylor-NOAA , feel free to merge this in if you agree it is ready.

PR #250, opened 2022-01-07
improve procedure O.O.O. to get variance correct

Rick-Methot-NOAA:


Rick-Methot-NOAA:
The run failure is now just a small difference in the se of two derived quantities for the simple_Lorenzen example, which failed to a greater degree before this change to the o.o.o. because this example requests se output for dynamic Bzero.
When I run this example on my local machine, I see the same difference. I also see that when 3.30.18 goes to the Hessian inversion step, it fails to invert on the first try, then adjusts the off-diagonals, and achieves an invertible Hessian.  The new code inverts on the first try, so I presume represents a better variance estimate.  Hopefully Cole's new console output will provide more transparency regarding what is going on at this stage.
inviting comment from @Cole-Monnahan-NOAA

Cole-Monnahan-NOAA:
@Rick-Methot-NOAA I'm not sure I totally follow this. ADMB has no internal procedure that adjusts the off-diagonals when doing the Hessian calculations, at least that I've seen in the source code.

The new output would certainly help pinpoint which stage it's in. It's currently broken for RE models but you can build this on [my fork](https://github.com/Cole-Monnahan-NOAA/admb) and run it to see. I've tested it on several SS models and it works fine.

Rick-Methot-NOAA:
So, what is ADMB doing when you see it go through 1 to several  displays of "Estimated covariance matrix my not be positive definite" followed by a vector of numbers, the same vector each time?

Cole-Monnahan-NOAA:
Those are in the delta method calculations of correlations. It appends the sdreport variables to the normal Hessian then inverts (I think). Those values are the eigenvalues, as seen here: https://github.com/admb-project/admb/blob/master/src/nh99/mod_sd.cpp#L172

You'll get the same vector because it's the same EVs from the same matrix. I don't fully understand what it's doing tp be clear. I do know that no one learns anything from those EVs so I turned them off. And printed which sdreport variable was causing that.

PR #255, opened 2022-01-22
Wip herma

Rick-Methot-NOAA:
@k-doering-NOAA 
@chantelwetzel-noaa 
@iantaylor-NOAA 
This corrects the biology_at_age_in_endyr output for hermaphroditism
herm_trans column is removed
replaced with sex_ratio calculated from numbers-at-age in endyr
ALSO!  the sex_ratio column will now appear in all 2 sex configurations, so will be new for r4ss
it is omitted in 1 sex models

### DIFF
--- a/SS_popdyn.tpl
+++ b/SS_popdyn.tpl
@@ -1336,6 +1336,7 @@ FUNCTION void get_time_series()
   //  SS_Label_Info_24.8  #hermaphroditism
       if(Hermaphro_Option!=0)
       {
+              warning<<y<<" "<<s<<" do herm;"<<endl;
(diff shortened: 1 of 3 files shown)


k-doering-NOAA:
I think the failing jobs can be ignored. The no-est job is failing because it assumes data_echo.ss_new is the name of the new file and it can't change the names properly to run the .ss_new model. 

The with-est job is failing because of small differences in Simple_Lorenzen; 

I suspect once this is merged into main, both of these jobs will pass.

Rick-Methot-NOAA:
Thanks Kathryn.  For this commit, the important check is to see if r4ss is happy with the new sex_ratio column.

Issue #228, opened 2021-11-17
add warning/exit for undefine/obsolete selectivity patterns


k-doering-NOAA:
Apparently, if pattern 30 is chosen for length selectivity, the model will run sometimes: [example](https://github.com/futureseas/SardineMSE/blob/1e17aa45ef118f53fe91f3914699bd256da47702/EM/EM_alldat/om.ctl#L220),even though according to the manual pattern 30 should no longer be used 

An explicit exit should be added to Stock Synthesis (and r4ss read function to prevent this)

@Rick-Methot-NOAA , I can work on this  if you think it is a good idea.

Rick-Methot-NOAA:
concur.  Please add a warning with recommendation on which better pattern to use.  Make it a warning with exit.

Issue #134, opened 2021-02-17
interpretation of "final" parameter in sex-specific selectivity offset with double normal


iantaylor-NOAA:
The Spiny Dogfish and Dover Sole assessments being conducted right now at NWFSC are both showing evidence of sex-specific selectivity (likely due to differences in distribution of the sexes for both species). Vlada and Chantel, leading these assessments, are exploring the use of the parameter offset option available for double-normal selectivity (set using a value of 3 or 4 in the "Male" column of the selex pattern input). This worked well for Big Skate in 2019 where the estimated selectivity was asymptotic. However, @chantelwetzel-noaa discovered some unusual patterns emerging when the "Scale" parameter is below the transformed "Final".

I replicated Chantel's patterns using the BigSkate_2019 model in the ss-examples repository (see figure below). The "final" = 0 for the main selectivity (males), which corresponds to selectivity = 0.5 and the "final" offset parameter is fixed at 0. The "scale" parameter is fixed at either 0.7 (left) or 0.3 (right).

It might be better to have the final value be scaled relative to the peak for each sex, rather than on the same 0 to 1 range. That would allow the scale parameter to scale down all elements of the curve equally without having to estimate an offset from final to achieve a similar shape, rather than allowing in somewhat strange patterns like the one on the right. If backwards compatibility is important, this could be new options 5 and 6, although the change would just be in interpretation of the offset, where any dome-shaped pattern should still be achievable through a slightly different parameter value.

Even if no change or addition is made to SS, it could be useful to expand the Selex24 spreadsheet to include the offset parameters.

![image](https://user-images.githubusercontent.com/4992918/108259970-97818880-7116-11eb-856d-5fc9015a57a4.png)

Rick-Methot-NOAA:
starting work on this now.  I suspect that I will create a new selectivity option rather than make a current one even more complicated with options

Rick-Methot-NOAA:
The combinations get complex very quickly.  I cannot imagine how to modify this approach without creating new, hard to test, situations with equally weird patterns.
I will start by copying current selex 24 and copying into currently discontinued selex 2, so we can use selex 2 as a testbed.
I really want to streamline the code for this selex 2.  There are TOO MANY "if" statements!  But the testing of any modification would be a big undertaking.

iantaylor-NOAA:
I'm sure there's a lot of complexity that I don't understand, but this is the sequence that I imagined:

[...]

PR #272, opened 2022-03-09
Issue 228 - remove and warn unused selex patterns

k-doering-NOAA:

#228 

This pull requests adds or edits warnings and exit statements for selectivity patterns that have been discontinued or are not recommended.

- [x] Link issue on pull request

I compiled and tested out all selectivity patterns that have exits and warnings or just warnings (length patterns 2, 4, 7, 30, 8; age patterns 13, 18).  I did this by modifying the selectivity patterns and number of lines required as in the manual (note pattern says it need 8 parameter lines, but only needs 6)

It would be good for Rick to review the code. In particular, if the single warnings when the model run is not supposed to exit on error are implemented correctly and if there are any changes in wording needed.

Someone checking at least a few of the patterns would be helpful, to make sure my tests also work elsewhere. 

- [ ] Changes in r4ss
- [x] Changes to SS3 manual
- [x] Changes to SSI (the SS3 GUI)
- [x] An entry in the stock synthesis change log (new features, bug reports)

### for ss3 manual

It might be good to remove the discontinued patterns from the manual; I think they have all been discontinued for years at this point. 

### for change log
Add warnings for discontinued selectivity patterns

### DIFF
--- a/SS_selex.tpl
+++ b/SS_selex.tpl
@@ -113,7 +113,7 @@ FUNCTION void get_selectivity()
           {
 			N_warn++;
 			cout << "EXIT - see warning" << endl;
-			warning << N_warn << " " << "Selectivity pattern 4 discontinued; use pattern 0 and special survey pattern 30 in data file instead."<<endl; 
+			warning << N_warn << " " << "Selectivity pattern 4 discontinued; use pattern 0 and special survey units 30 in data file instead."<<endl; 
 			exit(1); 
 			break;
 		  }
--- a/SS_selex.tpl
+++ b/SS_selex.tpl
@@ -88,15 +88,86 @@ FUNCTION void get_selectivity()
                 break;
               }
 
-  //  SS_Label_Info_22.3.2 #case 2 discontinued; use pattern 8 for double logistic
-           case 2:
-		   {
-             N_warn++; 
-			 cout<<" EXIT - see warning"<<endl; 
-			 warning<<N_warn<<" "<<"Selectivity pattern 2 discontinued; use pattern 8 instead."<<endl; 
-			 exit(1);
-             break;
-           }
+  //  SS_Label_Info_22.3.2 #case 2 size selectivity using double_normal_plateau and lots of bells and whistles; old version of 24 available for back compatibility
+            case 2:
+  //  SS_Label_Info_22.3.24 #case 24 size selectivity using double_normal_plateau and lots of bells and whistles
+            case 24:
+            {
+			  if(seltype(f,3)<3 || (gg==1 && seltype(f,3)==3) || (gg==2 && seltype(f,3)==4))
+				{peak=sp(1); upselex=mfexp(sp(3)); downselex=mfexp(sp(4)); final=sp(6); Apical_Selex=1.;}
+				else
+				{   // offset male parameters if seltype(f,3)==3, female parameters if seltype(f,3)==4
+				  peak=sp(1)+sp(Maleselparm(f));
+				  upselex=mfexp(sp(3)+sp(Maleselparm(f)+1));
+				  downselex=mfexp(sp(4)+sp(Maleselparm(f)+2));
+				  if(sp(6)>-999.) final=sp(6)+sp(Maleselparm(f)+3);
+				  Apical_Selex=sp(Maleselparm(f)+4);
+				}
+
+				if(sp(5)<-1000.)
+				{
+				  j1=-1001-int(value(sp(5)));      // selex is nil thru bin j1, so set sp(5) equal to first bin with selex (e.g. -1002 to start selex at bin 2)
+				  sel(1,j1)=1.0e-06;
+				}
+				else
+				{
+				  j1=startbin-1;                // start selex at bin equal to min sizecomp databin  (=j1+1)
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA:
I like the idea of removing the discontinued patterns from the User Manual. If someone has a very old model that includes one of these patterns, they can use a very old manual for information about it.

Rick-Methot-NOAA:
Did you check & update the list of selex fxns that displays in control.ss_new?

k-doering-NOAA:
> Did you check & update the list of selex fxns that displays in control.ss_new?

No, I didn't, good point! I can make that change.

Rick-Methot-NOAA (changes requested):
Let me try to hunt the history of:
  //  SS_Label_Info_22.3.2 #case 2 discontinued; use pattern 8 for double logistic
  //  SS_Label_Info_22.3.24 #case 24 size selectivity using double_normal_plateau and lots of bells and whistles
I suspect that I needed an experimental version of 24, so I copied the 24 code, renamed it as case 2.  So, I will do a code compare of the 24(2)  relative to the actual 24.

k-doering-NOAA:
Ok @Rick-Methot-NOAA, let me know what you find out!

Rick-Methot-NOAA:
all that needs to change are the comments labelling the code:
this line gets deleted:
// SS_Label_Info_22.3.2 #case 2 discontinued; use pattern 8 for double logistic
this line:
// SS_Label_Info_22.3.24 #case 24 size selectivity using double_normal_plateau and lots of bells
becomes:
// SS_Label_Info_22.3.2 #case 2 is a modification of case24 size selectivity for better performance with sex-specific

k-doering-NOAA:
OK. Actually, were you suggesting we keep two selectivity patterns,  or just the one with better performance (e.g., should actually 24 be deprecated and 2 kept)? Or maybe we just continue to say 2 is deprecated and add the small chunk of code you discovered to 24 to improve the pattern?

Rick-Methot-NOAA:
Keep two patterns for continued backward compatibility:
  case 24
 24 modified and currently with case 2 as its reference.

iantaylor-NOAA:
Linking in issue #134 for future reference, which explains the origins of the current selectivity pattern 2.

I like the idea of making 24 be the one with the modification, if I'm interpreting that suggestion correctly, and 2 be the backup for compatibility if we can document it adequately so that people know to look there.

This only matters for the small subset of models which had sex-specific double-normal selectivity which was estimated as dome-shaped.

Rick-Methot-NOAA:
I like Ian's suggestion.

k-doering-NOAA:
This does make sense, maybe we just have a note appear in the warning file if using 24?

It makes sense to have the improved version be added to 24. I will work on making these changes.

Rick-Methot-NOAA:
warning would be good:
IF using sex-specific version of 24, then warn user to consider better performance of case 2.

iantaylor-NOAA:
Warning could be conditional on this check for selex type 2 or 24 within the check for sex-specific selectivity: https://github.com/nmfs-stock-synthesis/stock-synthesis/blob/cdac3a738f01e2b9040546f908e38ea06e7994ea/SS_readcontrol_330.tpl#L3027-L3034

It seems like we're talking about two potential options:
1. **restore selex type 2** (removed in this pull request because we forgot to document that it has a purpose) **and leave selex type 24 as it stands with a warning** that selectivity type 2 could be used if you want the estimate of scale to apply to the final selectivity as well as the plateau, rather than just the plateau. This has the benefit of providing easy backwards compatibility and avoids explaining why models may have changed.
2. **restore selex type 2 and then swap the code for 2 and 24** so that type 2 matches the current type 24 where the scale parameter only applies to the plateau and not the final, and type 2 would be used for backwards compatibility. This has the benefit of bringing all the existing models to the "improved" parameterization automatically (if we agree that it's an improvement) without having to nudge people to explore that change. However, it has the drawback that some people may be surprised to see a model change and have to go through extra work to understand why.

I don't have any strong opinion on which is best.

Rick-Methot-NOAA:
Thanks for the clear description Ian.  Approach 1 would mean that a user wanting to keep the existing selex 24 for a sex-specific selectivity would get the warning every time until they switch to the new 2.
With approach 2, a user (there are very, very few!) of current sex-specific 24 would see a change in result and would not get a warning.  Then they write to us and we explain.  That seems OK.

k-doering-NOAA:
Ok, I went with option 2. I also realized the 2 patterns only had one line of difference, so combined the code for both in the switch statement and added an if/else statement.

I used tested the change using a modified version of big skate (had to change parameter 6 from -999 to 0.01 test the code). I also modified it 1 additional time to add pattern 2 instead of 24 and checked with print debugging that SS3 was going to the correct if statements.

This same test could be done, looking for a warning message when using pattern 2 and none when using pattern 24; I imagine there would be a difference in the selectivity patterns when using these.

Rick-Methot-NOAA:
I like your approach to coding this.  I think it is ready to merge, then get documented in the change log and the manual and SSI.

Rick-Methot-NOAA (commented):
all good.  merge OK

PR #276, opened 2022-03-21
Issue 232 added lines to rebuild.dat

nschindler-noaa:
Additional lines for West Coast rebuilder

- [x] Link issue on pull request

- [ ] Changes in r4ss
- [x] Changes to SS3 manual
- [x] Changes to SSI (the SS3 GUI)
- [ ] An entry in the stock synthesis change log (new features, bug reports)

### DIFF
--- a/SS_readdata_330.tpl
+++ b/SS_readdata_330.tpl
@@ -3268,7 +3268,7 @@
   	{N_warn++; warning<<N_warn<<"; changing Imple_Error to 0 because no forecast "<<endl; Impl_Error_Std=0.0; Do_Impl_Error=0;}
   }
 
-  echoinput<<endl<<"#next select rebuilding program output (0/1)"<<endl;
+  echoinput<<endl<<"#next select rebuilding program output: 0=none; 1=standard; 2=expanded"<<endl;
   *(ad_comm::global_datafile) >> Do_Rebuilder;
   echoinput<<Do_Rebuilder<<" # echoed value"<<endl;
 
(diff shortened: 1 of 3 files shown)


Rick-Methot-NOAA (approved):
These changes are approved.

chantelwetzel-noaa:
I was just looking over the changes to update the manual and if I am understanding the options, I see some potential issues with the options as currently coded. Now there are 3 options in the forecast file for the "Do Rebuilder": 0=no rebuild.dat file created, 1=create rebuild.dat file, and 2=rebuild.dat file with additional inputs. In the SS_write.tpl file (lines 749-756 and 766-770) if looks like these new required input lines are only written to the rebuild.dat file if the Do Rebuilder option is set to 2. However, these new parameter lines should be written with option 1 or option 2 for the "Do_Rebuilder". I think the only difference should have been that if the Do Rebuilder line in the forecast file is set to option 2 additional inputs in the forecast file would be needed which would populate the values in these required rebuild.dat parameter lines.  

@Rick-Methot-NOAA Can you look at the SS_write.tpl file to see if my interpretation is correct?

Rick-Methot-NOAA:
option 2 is there so that option 1 provides backward compatibility.  If there is no need for backward compatibility, then option 2 can be removed and the expanded output will be the only option.

Rick-Methot-NOAA:
@nschindler-noaa   Can you make this change per Chantel's advice?

k-doering-NOAA:
Chantel, I think part of your observations was that additional inputs to the forecast file are expected when using rebuilder - is this the case? I'm not seeing changes to the code to read in extra inputs.

Rick-Methot-NOAA:
There is no change to the forecast.ss inputs.  All that is changed with this option is that some additional outputs are provided.  It is then up to the user of rebuild.dat to edit those additional values as needed.  So, I think it makes sense to simply make this expanded output the only choice.

PR #277, opened 2022-03-23
create skeleton for MV_Tweedie

Rick-Methot-NOAA:
## What issue(s) does this PR address? Describe and add issue numbers, if applicable.

- [ #266 ] Link issue on pull request

## What tests have been done? Upload any model input files created for testing in a zip file, if possible.
none
## What tests/review still need to be done? Who can do it, and by when is it needed (ideally)?
everything
## Check which is true. This PR requires:

- [x] Changes in r4ss
- [x] Changes to SS3 manual
- [x] Changes to SSI (the SS3 GUI)
- [x] An entry in the stock synthesis change log (new features, bug reports)

PR #278, opened 2022-03-24
Mv tweedie complike

James-Thorson-NOAA:
Adding mvTweedie likelihood, using `dtweedie` in ADMB-14

- [#277] Link issue on pull request

No tests done!

1. Adding input/output features to get mvTweedie parameters
2. Checking `dtweedie` exists in ADMB
3. Checking likelihood calculation
4. Add code for effective-sample-size calculation somewhere TBD
5. testing on simple example and compare with VAST

- [ ] Changes in r4ss
- [X] Changes to SS3 manual

##  Describe any changes in r4ss/SS3 manual/SSI/change log that are needed (if checked):

to discuss later I think

### DIFF
--- a/SS_objfunc.tpl
+++ b/SS_objfunc.tpl
@@ -290,15 +290,19 @@ FUNCTION void evaluate_the_objective_function()
        if(gen_l(f,i) >=2 && gender==2) length_like(f,i) -= nsamp_l(f,i) *
        obs_l(f,i)(tails_w(3),tails_w(4)) * log(exp_l(f,i)(tails_w(3),tails_w(4)));
       }
+<<<<<<< Updated upstream
       else if(Comp_Err_L(f)==1)  //  dirichlet
+=======
+      if( (Comp_Err_L(f)==1) | (Comp_Err_L(f)==2) ) //  dirichlet
+>>>>>>> Stashed changes
       {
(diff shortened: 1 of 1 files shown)


SS_objfunc.tpl
```diff
@@ -321,9 +321,29 @@ FUNCTION void evaluate_the_objective_function()
         }
         length_like(f,i)-=temp;
       }
-      else   //  MV_Tweedie
+      else  //  multivariate-Tweedie
```
iantaylor-NOAA, line 324: I think this we should just assign 3 to this new option and changed this line to 
`if( (Comp_Err_L(f)==3)`
otherwise I think multinomial likelihood will trigger both the section above with
`if( (Comp_Err_L(f)==0)`
and also this new section

James-Thorson-NOAA: makes sense!  Thanks for catching that.  I can't compile and run it without parameter declarations, so bound to be a few issues like this.  

did you make the change in the PR code?  I'm not used to code reviews during PR, but seems great :)

iantaylor-NOAA: No, just added a comment on the code change (by clicking on line under "Files changed" under pull request on github.com, but I haven't made any change yet. I think that I could either push a change to the new branch on your fork to be included in this PR, or we could merge your PR into the branch on the nmfs-stock-synthesis version and then I can push the change to the branch there. Since nobody is using the branch for anything but this likelihood, it makes sense to merge first.

Issue #280, opened 2022-03-28
tag-recapture: create user filter to set min number of recaptures for inclusion of a tag-group


Rick-Methot-NOAA:
some tag groups may have many releases, but few recaptures.  It seems useful to allow a user to set a filter such that that number of recaptures in time >= mixperiod exist.
This could be implemented in ss_readdata330.tpl to create a "use" flag for each tag group.
The value could be read in ss_readdata330.tpl right after the input of the value for mixperiod.
The total number of recaps is currently not counted, but easily could be in ss_readdata330.tpl.
Note: remember to only count those that are >=mixperiod.
Probably should make this a conditional input:
if(mixperiod>=0){
min_tag_recap=0.0;}
else{
mixperiod=-mixperiod;
ad_comm:  min_tag_recap;}

Rick-Methot-NOAA:
1 # do tags (0/1)
121 # N tag groups
769 # N recap events
-4 # TG_mixperiod (latency) is the first period in which logL is calculated for this tag group (0 = release period)
#  (use negative value for TG_mixperiod to invoke reading TG_min_recap after TG_maxperiods
16 # max periods (seasons) to track recoveries, after which tags are still read into this accumulator period
2 # TG_min_recap:  N recaps >= mixperiod must be >= min_recap to include tag group in logL
# Release data for each tag group.  Tags are considered to be released at the beginning of a season (period)

ALSO*****
In compreport.sso, I now output the time in periods since release, which seems more useful for plotting and comparisons than real time, which is still there in another column

k-doering-NOAA:
The final output looks like this:

```
2 # do tags (0/1/2). 2 means include TG_min_recap input
121 # N tag groups
769 # N recap events
4 # TG_mixperiod (latency) is the first period in which logL is calculated for this tag group (0 = release period)
16 # max periods (seasons) to track recoveries, after which tags are still read into this accumulator period
2 # Cond on Do Tags 2: TG_min_recap:  N recaps >= mixperiod must be >= min_recap to include tag group in logL
# Release data for each tag group.  Tags are considered to be released at the beginning of a season (period)
```

PR #281, opened 2022-03-29
complete addition of TG_min_recap, and revise compreport

Rick-Methot-NOAA:
This adds capability to exclude from the tag logL tag groups with less than a specified minimum number of recaptures
- [#280  ] Link issue on pull request

works with the IOTC Yellowfin tuna spatial model
none

- [X] Changes in r4ss
- [X] Changes to SS3 manual
- [X] Changes to SSI (the SS3 GUI)

##  Describe any changes in r4ss/SS3 manual/SSI/change log that are needed (if checked):
input of new feature is conditional on TG_mixperiod having a negative value:
1 # do tags (0/1)
121 # N tag groups
769 # N recap events
-4 # TG_mixperiod (latency) is the first period in which logL is calculated for this tag group (0 = release period)
#  (use negative value for TG_mixperiod to invoke reading TG_min_recap after TG_maxperiods
16 # max periods (seasons) to track recoveries, after which tags are still read into this accumulator period
2 # TG_min_recap:  N recaps >= mixperiod must be >= min_recap to include tag group in logL

## Additional information (optional):
@jyqalan

### DIFF
--- a/SS_write_ssnew.tpl
+++ b/SS_write_ssnew.tpl
@@ -335,7 +335,7 @@ FUNCTION void write_nudata()
       report1<<TG_min_recap<< " # TG_min_recap ## tag group's N recaps >=mixperiod must be >= min_recap to include tag group in logL)"<<endl;
     }
     else{
-    report1<< " # COND:  TG_min_recap ##  use Do_TG=2 to invoke reading TG_min_recap after TG_maxperiods"<<endl;
+    report1<< "# COND:  TG_min_recap ##  use Do_TG=2 to invoke reading TG_min_recap after TG_maxperiods"<<endl;
     }
 
     // tag releases
--- a/SS_write_ssnew.tpl
+++ b/SS_write_ssnew.tpl
@@ -329,8 +329,15 @@ FUNCTION void write_nudata()
     // info on dimensions of tagging data
     report1<<N_TG<<" # N tag groups"<<endl;
     report1<<N_TG_recap<<" # N recap events"<<endl;
-    report1<<TG_mixperiod<<" # mixing latency period: N periods to delay before comparing observed to expected recoveries (0 = release period)"<<endl;
-    report1<<TG_maxperiods<<" # max periods (seasons) to track recoveries, after which tags enter accumulator"<<endl;
+    report1<<TG_mixperiod_rd<<" # TG_mixperiod (latency) is the first period in which logL is calculated for this tag group (0 = release period)"<<endl;
+    report1<<"#  (use negative value for TG_mixperiod to invoke reading TG_min_recap after TG_maxperiods"<<endl;
+    report1<<TG_maxperiods<<" # max periods (seasons) to track recoveries, after which tags are still read into this accumulator period"<<endl;
+    if(TG_mixperiod_rd>=0){
+      report1<<"#  TG_min_recap would go here if TG_mixperiod < 0"<<endl;
+    }
+    else{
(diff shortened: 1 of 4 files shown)


k-doering-NOAA:
@Rick-Methot-NOAA, did you want this to be part of the 3.30.19?

Rick-Methot-NOAA:
Yes because we still are waiting on the rebuilder revision.
Will need some work on the manual and r4ss.

SS_write_ssnew.tpl
```diff
@@ -329,8 +329,15 @@ FUNCTION void write_nudata()
     // info on dimensions of tagging data
     report1<<N_TG<<" # N tag groups"<<endl;
     report1<<N_TG_recap<<" # N recap events"<<endl;
-    report1<<TG_mixperiod<<" # mixing latency period: N periods to delay before comparing observed to expected recoveries (0 = release period)"<<endl;
-    report1<<TG_maxperiods<<" # max periods (seasons) to track recoveries, after which tags enter accumulator"<<endl;
+    report1<<TG_mixperiod_rd<<" # TG_mixperiod (latency) is the first period in which logL is calculated for this tag group (0 = release period)"<<endl;
+    report1<<"#  (use negative value for TG_mixperiod to invoke reading TG_min_recap after TG_maxperiods"<<endl;
+    report1<<TG_maxperiods<<" # max periods (seasons) to track recoveries, after which tags are still read into this accumulator period"<<endl;
+    if(TG_mixperiod_rd>=0){
+      report1<<"#  TG_min_recap would go here if TG_mixperiod < 0"<<endl;
```
k-doering-NOAA, line 336: Maybe reword to `"# COND: TG_min_recap if TG_mixperiod < 0"` to be more consistent in style with the manual.

SS_write_ssnew.tpl
```diff
@@ -329,8 +329,15 @@ FUNCTION void write_nudata()
...
+    }
+    else{
+      report1<<TG_min_recap<<" # TG_min_recap:  N recaps >= mixperiod must be >= min_recap to include tag group in logL"<<endl;
```
k-doering-NOAA, line 339: Is there a way to reword this more concisely? I am having trouble following the statements. Maybe just simpl put 
`# COND: TG_min_recap if TG_mixperiod < 0` to mirror the statement produced when the mix period is greater than 0?

k-doering-NOAA (approved):
Looks good! I took a look at the code, then Rick walked me through the changes while screen sharing and explained why they were made. Makes sense to me.

Optional changes:
- ssnew comments for the new conditional input
- using a new 2 option for the do tag flag as the trigger for the conditional input rather than a negative value on the mix period. Rick made the observation that this may be better and I agree it may be more intuitive for users.

Rick-Methot-NOAA:
revise trigger
2 # do tags (0/1/2); where 2 allows entry of TG_min_recap
instead of negative mixperiod value as trigger.

k-doering-NOAA:
Looks good @Rick-Methot-NOAA ! I ran a version of the YFT for the spatial project, and it worked well. Attached are the input files for the version with the min recapture option for future.

Running with `-stopph 0 -nohess`, only the 2 tag likelihoods changed when using the min recapture option, and the likelihoods were reduced, so this checks out.

I also tried running with estimation as well, and the model ran fine.

Something picky: I noticed there was a space in the .ss_new file when printing the # COND line; I will push a small change to get rid of it. 

This is ready to merge in once I push the picky change.

[use_min_option_inputfiles.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/8389962/use_min_option_inputfiles.zip)

PR #284, opened 2022-04-05
Fix seasonal recruitment

### LINKED ISSUE #283: errors in seasonal recruitment and empirical wtatage
Two issues were detected while working with Peter Kuriyama's anchovy  model.  One related to use of 2 seasons, and one related to use of empirical wtatage.
1.  recruitment in the Time_Series table is recruitment by area, not total recruitment.  They should be the same in a one area model.  However, in a 2 season model with all recruits assigned to one season, a double counting occurs as the code loops seasons.  The recruitment reported in this table gets accumulated for each season.  
FIX by making the accumulation conditional on Settle_season.  This was the case several releases ago, but got changed.
2.  Within the report.sso sequence to output biology, there were insufficient traps to prevent biology based fecundity calculations from overwriting fecundity as read from wtatage.ss.  This occurred late in the output so few other items were affected by this overwriting.  It was noticed when looking at SPR/YPR profile and the global MSY would also be affected.  This issue may have caused when wtage storage was reconfigured in 3.30.17 or 0.18
   FIX by adding appropriate conditional statements like:  
            if(WTage_rd==0) get_mat_fec();

Rick-Methot-NOAA:

- [#283  ] Link issue on pull request

compare output between tables for the anchovy assessment files
another sharp set of eyes

- [ ] Changes in r4ss
- [ ] Changes to SS3 manual

### DIFF
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -3513,12 +3513,9 @@ FUNCTION void SPR_profile()
         Do_Equil_Calc(equ_Recr);
 //  SPAWN-RECR:   calc equil spawn-recr in the SPR loop
         SPR_temp=SSB_equil;
-        SS2out<<"unfished S and R: "<<SSB_unf<<" "<<Recr_unf<<" SSB/R: "<<SPR_temp<<" fec: "<<fec<<" ";
-//        Equ_SpawnRecr_Result = Equil_Spawn_Recr_Fxn(SR_parm(2), SR_parm(3), SSB_virgin, Recr_virgin, SPR_temp);  //  returns 2 element vector containing equilibrium biomass and recruitment at this SPR
         Equ_SpawnRecr_Result = Equil_Spawn_Recr_Fxn(SR_parm_work(2), SR_parm_work(3), SSB_unf, Recr_unf, SPR_temp);  //  returns 2 element vector containing equilibrium biomass and recruitment at this SPR
         Btgt_prof=Equ_SpawnRecr_Result(1);
         Btgt_prof_rec=Equ_SpawnRecr_Result(2);
-        SS2out<<"fished S and R: "<<Btgt_prof<<" "<<Btgt_prof_rec<<" SSB/R: "<<Btgt_prof/Btgt_prof_rec<<endl;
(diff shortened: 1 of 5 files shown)


SS_write_report.tpl
```diff
@@ -2546,7 +2546,7 @@ FUNCTION void write_bigoutput()
               get_growth3(styr,t,s, subseas);
               Make_AgeLength_Key(s, subseas);  //  spawn subseas
             }
-            get_mat_fec();
+            if(WTage_rd==0) get_mat_fec();
```
k-doering-NOAA, line 2549: @Rick-Methot-NOAA , can you explain this line to me? Was this a separate issue from the seasonal recruitment or part of the same?

Rick-Methot-NOAA:
This second issue was detected while diagnosing the same set of input files. It is due to the non-modularity of the growth&biology parameters vs. the input of all biology from wtatage.ss   The problem was that fecundity, as calculated from growth parameters, was erroneously overwriting the fec(g) array.  Really should have been a separate issue with separate fix.

Rick-Methot-NOAA:
r4ss problems probably due to some extra output lines that are not finalized.

k-doering-NOAA:
> r4ss problems probably due to some extra output lines that are not finalized.

Yes, that makes sense. We should probably get those sorted out before merging into main, right?

Here is the error: https://github.com/nmfs-stock-synthesis/stock-synthesis/runs/5841949938?check_suite_focus=true#step:12:72

```
<simpleError in order(yielddat[["Depletion"]], decreasing = FALSE): argument 1 is not a vector>
```

k-doering-NOAA:
> This second issue was detected while diagnosing the same set of input files. It is due to the non-modularity of the growth&biology parameters vs. the input of all biology from wtatage.ss The problem was that fecundity, as calculated from growth parameters, was erroneously overwriting the fec(g) array. Really should have been a separate issue with separate fix.

Thanks for explaining!

Rick-Methot-NOAA:
with nearly every check passed, I will let this issue sit until I return to office on 4/11,

PR #285, opened 2022-04-06
refine pull request template

iantaylor-NOAA:

Changes the PR template checklist so checked boxes mean a complete rather than incomplete task. 
Status-quo template makes it look like 4 checked boxes doesn't require more work.
image

- [ ] Changes in r4ss
- [ ] Changes to SS3 manual
- [ ] An entry in the stock synthesis change log (new features, bug reports)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,7 @@
 
 ## What issue(s) does this PR address? Describe and add issue numbers, if applicable.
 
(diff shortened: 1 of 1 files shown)


k-doering-NOAA:
Thanks @iantaylor-NOAA , this is a good point! I also pushed a change to remove an unnecessary checkbox.

k-doering-NOAA (approved):

Issue #298, opened 2022-04-25
Fspr nan values when running without bmark/fcast and with hessian


k-doering-NOAA:
I turned off benchmark and forecast for this [two morphs model](https://github.com/nmfs-stock-synthesis/test-models/tree/two_morph/models/two_morph) and when running with hessian with ss3.30.19, I get an error in the forecast:

```
Estimating row 86 out of 86 for hessian
6 706 -log(L): 2643.81  Spbio: 99155.3 11415.8
6 707 -log(L): 2643.81  Spbio: 99155.3 11415.8
6 708 -log(L): 2643.81  Spbio: 99155.3 11415.8
6 709 -log(L): 2643.81  Spbio: 99155.3 11415.8
do benchmark and forecast if requested in sdphase
 got Fspr nan nan
array bound exceeded -- index too high in prevariable::operator()
```

this doesn't occur when I run without estimating the hessian. I'm thinking this may be a bug?

I haven't tried testing with other models, but can if it would be helpful.

Rick-Methot-NOAA:
I suspect the code lost a conditional check for benchmark while several related items were moved around.  Should be an easy fix near bottom of ss_proced.tpl

k-doering-NOAA:
Thanks @Rick-Methot-NOAA , I think your suspicion is correct! I submitted PR #300 for your review with what I think is the fix.

PR #300, opened 2022-04-26
fix, #298: add conditional statement around bmark code

k-doering-NOAA:

Link issue(s) here: #298 

I tested with the two morphs model (attached), compiled SS3 locally, and ran with no options. The nan message disappeared, however there was a covariance matrix may not be positive definite warning. I'm not sure if this has something to do with the SS3 code, or if it is specific to the model.

[two_morph.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/8563411/two_morph.zip)

If @Rick-Methot-NOAA could take a look and make sure this fix does what I think it does, that would be great!

Add line in the change log about this bug.

### DIFF
--- a/SS_popdyn.tpl
+++ b/SS_popdyn.tpl
@@ -344,16 +344,15 @@ FUNCTION void get_initial_conditions()
     SSB_virgin = SSB_equil;
     SPR_virgin = SSB_equil / Recr_virgin; //  spawners per recruit
   
-    //  unnecessary because these are now done in benchmark itself
-    //    if(Do_Benchmark==0)
-    //    {
-    //      Mgmt_quant(1)=SSB_virgin;
-    //      SSB_unf=SSB_virgin;
-    //      Recr_unf=Recr_virgin;
-    //      Mgmt_quant(2)=totbio;  //  from equil calcs
-    //      Mgmt_quant(3)=smrybio;  //  from equil calcs
-    //      Mgmt_quant(4)=Recr_virgin;
-    //    }
+        if(Do_Benchmark==0)
+        {
+          Mgmt_quant(1)=SSB_virgin;
+          SSB_unf=SSB_virgin;
+          Recr_unf=Recr_virgin;
+          Mgmt_quant(2)=totbio;  //  from equil calcs
+          Mgmt_quant(3)=smrybio;  //  from equil calcs
+          Mgmt_quant(4)=Recr_virgin;
+        }
   
     Smry_Table(styr - 2, 1) = totbio; //  from equil calcs
     Smry_Table(styr - 2, 2) = smrybio; //  from equil calcs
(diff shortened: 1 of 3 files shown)


k-doering-NOAA:
I turned back on benchmarks, but left forecasts off for this model, and there was no warning about the covariance matrix. So I am wondering if there are additional changes in SS3 that need to be made to get rid of this warning when both bmarks and forecasts are off.

Rick-Methot-NOAA:
That change is exactly what I expected.
Please compare ss_summary.sso with and without benchmark turned on.  Without benchmark, the list of management quantities is much shorter, but cannot be of zero length.  So, mgmt_quant(1) should get assigned ssb_virgin if not doing benchmark.  I do not see that in the code.  Should look back a few versions to see when it disappeared.

k-doering-NOAA:
Good call about looking at ss_summary.sso with and without benchmark. In the management quanties when benchmark is off, this is what I see: 
```
#_Mgmt_Quantity
Bzero_again 0 0
Totbio_unfished 0 0
SmryBio_unfished 0 0					
Recr_unfished 0 0							
```

Is this ok? OR do you think it would be more clear to eliminate some of these?

The only other difference were the Bratio value, which were present but quite different when bmark was turned off ( think this is because the depletion basis was different). Maybe they should not be showing up?

Rick-Methot-NOAA:
Bzero_again is labelled as distinct reminder that it is there as a placeholder so that some value in mgmt_quantity will have a value and that value will have variance estimated.  It is the zero variance around the zero value of bzero_again that causes the hessian to not be posdef.

So, these lines in benchmark:
  Mgmt_quant (1) = SSB_unf;
  Mgmt_quant (2) = totbio;
  Mgmt_quant (3) = smrybio;
  Mgmt_quant (4) = Recr_unf;

need to be executed even when do_benchmark==0.

I looked in the ss.tpl for 3.30.14 and found:
      if(Do_Benchmark>0)
      {
        Get_Benchmarks(show_MSY);
      }
      else
      {Mgmt_quant(1)=SSB_virgin;  warning<<"set to virgin in proced if no benchmark"<<endl;}

That clause disappeared when I was doing some house-cleaning.

Note that I do not think the warning is needed.

k-doering-NOAA:
Thanks, Rick! Your detective work was really helpful in figuring this out.

I found some commented out lines that I suspect were turned off at the same time; I uncommented them in [61e63c9](https://github.com/nmfs-stock-synthesis/stock-synthesis/pull/300/commits/61e63c9ff13af238844fd548d002d4c06616441a) and this assigned values to to the 4 management quantities when bmark is off, thus preventing the non-posdef covariance matrix error.

I'm not sure if this is the cleanest way to do this, perhaps these lines should be moved to another section of the code?

Rick-Methot-NOAA:
I think your solution was the correct one.

k-doering-NOAA:
Thanks, Rick, I think this good to merge when you and Neal are ready (given the formatting work)

The failing call-test-r4ss-with-ss3 job is an issue that has been fixed in https://github.com/r4ss/r4ss/pull/659 , so I think it can be safely ignored.

Rick-Methot-NOAA (approved):
good collaboration on this one.  Thanks for fixing the problem.

PR #307, opened 2022-05-05
Add pred m to z not m

Rick-Methot-NOAA:
## What issue(s) does this PR address? Describe and add issue numbers, if applicable.

Link issue(s) here: #88 and #235 

## What tests have been done? Upload any model input files created for testing in a zip file, if possible.
extensive look at all test models, special testing using the predator model:  vendance.
## Check which is true. This PR requires:

- [ ] no further changes to r4ss
- [ ] no further changes to the manual
- [ ] no further changes to SSI (the SS3 GUI)
- [ ] no further changes to the stock synthesis change log (new features, bug reports)

## Describe any changes in r4ss/SS3 manual/SSI/change log that are needed (if not checked):
natural mortality report table has been reformatted
technical description of these changes is in attached doc
[natM tech doc.txt](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/8635992/natM.tech.doc.txt)

PR #310, opened 2022-05-10
fix forecast M with predators

Rick-Methot-NOAA:

Link issue(s) here: 
fix to main to repair M in forecast when predators are used.  Note that a separate development branch changes the pred_M2 approach and is not affected by this problem

detailed test using vendance 
none

- [ x] no further changes to r4ss
- [ x] no further changes to the manual
- [x ] no further changes to SSI (the SS3 GUI)

this must get noted in the change log

### DIFF
--- a/SS_benchfore.tpl
+++ b/SS_benchfore.tpl
@@ -2204,8 +2204,10 @@ FUNCTION void Get_Forecast()
         ALK_subseas_update = 1;
         get_growth2(y);
       }
-      if (timevary_MG(y, 1) > 0 || N_pred>0)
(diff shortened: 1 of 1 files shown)


Rick-Methot-NOAA:
The change will not be needed after we merge in #307, but a one line fix to the main branch while waiting for a hundreds of lines development branch seemed like a good idea.

k-doering-NOAA (approved):
Thanks, Rick. That makes sense.

Rick-Methot-NOAA:
I do not think we need to release this as a formal fix as it affects only the predation applications.  But it is good to have in case it is needed.

Issue #314, opened 2022-05-16
Linux build script


arni-magnusson:
Hi, I wanted to build SS3 on Linux and found the Make_SS_330_new.sh script to have the following small issues:

1. On my machines (Debian 11 and Ubuntu 22.04), the script will not start and returns an error:
```
./Make_SS_330_new.sh: 6: Syntax error: "(" unexpected
```
2. I tried running Make_SS_330_new.sh a few times before understanding that either the 'opt' or 'safe' build must be selected. It would be nice if the --help message mentioned that.

3. The --help message starts with a warning:
```
line 86: [0+1==1]: command not found
```

4. It would be nice (and harmless) if the build script creates the build directory, rather then returning an error that the build directory does not exist.
```
line 129: cd: SS330: No such file or directory
```
The automatic creation of a build directory can be considered standard behavior in build scripts.

5. The --opt flag is essentially ignored by the script. When run with --opt, it produces a safe (not optimized) build:
```
ADMB Program: ss_opt
ADMB-12.3 safe libraries was compiled using GNU C++ 10.2.1 64bit on Sep 23 2021
```

I have prepared a pull request that fixes the above issues.

Rick-Methot-NOAA:
Hi Arni.  Thanks for helping to improve SS3.  Kathryn will be back tomorrow and look into these suggestions.

PR #315, opened 2022-05-16
Linux build

arni-magnusson:

Link issue(s) here: #314

Built SS3 for Linux locally

None

- [ x ] no further changes to r4ss
- [ x ] no further changes to the manual
- [ x ] no further changes to SSI (the SS3 GUI)
- [ x ] no further changes to the stock synthesis change log (new features, bug reports)

None

### DIFF
--- a/.github/workflows/call-build-ss3.yml
+++ b/.github/workflows/call-build-ss3.yml
@@ -3,6 +3,7 @@ on:
   push:
     paths:
     - '**.tpl'
+    - '**.sh'
   workflow_dispatch:
 jobs:
   call-workflow:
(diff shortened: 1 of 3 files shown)


k-doering-NOAA:
Thanks so much for this PR, @arni-magnusson! To my knowledge, only our SS3 team has used it, so it is great to make it more general.

 @nschindler-noaa, would you mind reviewing, since you put together this build script?

I would also like to make sure we can still build with  github actions before merging in, but I think we need to change some settings in our repository (which I don't have permissions to do) to run the actions on a fork. Hopefully will do this later today!

Issue #321, opened 2022-05-23
Is abort() different than exit(1)?


k-doering-NOAA:
I see that the `abort()` function is used [when ss3 doesn't read the ctl file successfully](https://github.com/nmfs-stock-synthesis/stock-synthesis/blob/36768a467ebd7e164cff57cc6ae734ddd0cb8c0e/SS_readcontrol_330.tpl#L4641) instead of `exit(1)` (which is used elsewhere throughout the ss3 source code). Is there a reason why? I couldn't tell from this [definition](https://www.cplusplus.com/reference/cstdlib/abort/)

Edit: apologize for the flood of issues, they are coming up as I work on the console output.

k-doering-NOAA:
Ah, this stack overflow post [explains it](https://stackoverflow.com/questions/397075/what-is-the-difference-between-exit-and-abort)

It seems rather technical, and we would be fine just using exit().

PR #323, opened 2022-05-24
#321: use exit(1) instead of abort()

k-doering-NOAA:

Link issue(s) here: #321

None, because this change is very simple. Given we use exit(1) elsewhere in the code, it doesn't seem necessary to test this small change.

@Rick-Methot-NOAA to review the change and "approve" it or request edits.

Not sure if this change is worth documenting in the change log? It's really small.

### DIFF
--- a/SS_readcontrol_330.tpl
+++ b/SS_readcontrol_330.tpl
@@ -4637,8 +4637,8 @@
   }
   else
   {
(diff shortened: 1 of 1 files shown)


Rick-Methot-NOAA (approved):
looks good

Issue #123, opened 2021-01-27
switch to compiling with ADMB 12.3 upon its release


k-doering-NOAA:
ADMB 12.3 will likely be released soon (probably in the next few weeks). Should we test it out once available and try to switch over to compiling SS with this version? I think there would be several advantages, including a fix that should speed up runs (relative to ADMB 12.2), the -hess_step option, and a fix for reading input files with OSX9 line endings.

k-doering-NOAA:
ADMB 12.3 has been released: https://github.com/admb-project/admb/releases/tag/ADMB-12.3 so we can start transitioning.

k-doering-NOAA:
@iantaylor-NOAA , would you mind asking IT to get admb 12.3 onto tumnus and mac mini? not too much a rush, given we have gh actions, but we may still need it sometimes.

I've update all the builds on gh actions (currently in a [branch](https://github.com/nmfs-stock-synthesis/stock-synthesis/tree/test_admb_12_3)) and am now testing that admb 12.3 works as expect on jenkins. We should know tomorrow if it does, and can merge in the branch at that time.

iantaylor-NOAA:
Request submitted for Tumnus (linux server at NWFSC).
Installing on the mac mini shared by a few of us at NWFSC created some errors noted at https://github.com/admb-project/admb/issues/197.

iantaylor-NOAA:
ADMB 12.3 has now been successfully installed on both Tumnus and the Mac mini.

kellijohnson-NOAA:
For what it is worth, i was able to successfully compile SS with the new version.

k-doering-NOAA:
Thanks so much @iantaylor-NOAA, I appreciate the help! And thanks Kelli, that is good to know; it also compiled on my machine.

@Rick-Methot-NOAA , have you been able to upgrade? I think we can close this issue once you have upgraded and pull request #144 is merged in, then just used 12.3 to compile from now on.

Rick-Methot-NOAA:
I have switched.  So we are good to go

PR #326, opened 2022-05-27
Fix: remove warnings coming from ss3 source code. Details below

k-doering-NOAA:
Issue was not created, but while working on #123, it became clear to me that some of the warning messages coming from SS3 could be taken care of. This will make it easier to identify new/different warnings in the future using github actions, and also cleans up the codebase a bit. In particular, the changes made are:

- add fallthrough statement to note implicit fallthrough is intentional
- get rid of maybe uninit warning
- use multiline comment instead of single line comment to get rid of warning
- remove parm_adjust_method from create_timevary because unused
- refactor: clean up comments for time varying so only in 1 place (no warning but found while getting rid of parm_adjust_method)
- try to fix warning re using wrong type in for statements
- use int instead of unsigned
- remove unused variable bio_t_base_from function

Link issue(s) here: n/a

periodically, pushed changes to github and checked that the change got rid of the warning using github actions run (note: squashed commits to simplify the PR).

@Rick-Methot-NOAA should check that none of the changes to fix the warnings are problematic.

Also, Check with the github action that the only remaining warnings are marked as [-Wdeprecated-copy] remain. 

I think some work that needs to be done before merging this in is revising the github action (probably in a branch in the workflow repository), so that it can do this same check automatically. The number of warnings expected will also be less than previously expected. I can work on this once Rick verifies the changes are not problematic.

### DIFF
--- a/SS_selex.tpl
+++ b/SS_selex.tpl
@@ -254,6 +254,7 @@ FUNCTION void get_selectivity()
               // #43 non-parametric size selex scaled by average of values at low bin through high bin
               case 43:
                 scaling_offset = 2;
+                [[fallthrough]];
               case 6:
               {
                 lastsel = -10.0; // log(selex) for first bin;
@@ -600,6 +601,7 @@ FUNCTION void get_selectivity()
               /*  uses max(raw vector) to achieve scale to 1.0 */
               case 42:
                 scaling_offset = 2;
+                [[fallthrough]];
               case 27:
               {
                 int j2;
@@ -1139,6 +1141,7 @@ FUNCTION void get_selectivity()
               //    transformation as selex=exp(parm); some special codes */
               case 41:
                 scaling_offset = 2;
+                [[fallthrough]];
               case 17: //
               {
                 lastsel = 0.0; //  value is the change in log(selex);  this is the reference value for age 0
@@ -1533,6 +1536,7 @@ FUNCTION void get_selectivity()
               // #42 cubic spline scaled by average of values at low age through high age
               case 42:
                 scaling_offset = 2;
+                [[fallthrough]];
               case 27:
               {
                 // define vectors which form the basis for cubic spline selectivity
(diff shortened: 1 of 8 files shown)


SS_selex.tpl
```diff
@@ -1,7 +1,7 @@
 // SS_Label_file  #11. **SS_selex.tpl**
 // SS_Label_file  # * <u>get_selectivity()</u>  // does length and age selectivity and retention for all fleets
 // SS_Label_file  # * <u>make_fish_selex()</u>  // for all fleets in a particular season, does the dot product of length distribution with length selectivity and retention vectors to calculate equivalent mean quantities at age for each platoon
-
+
```
k-doering-NOAA, line 4: something weird is going on with the line endings. I will try to fix.

k-doering-NOAA: Github told me when I tried to edit inline that this file had a mix of line endings, which is why this is happening. It is changing them all to windows style (CRLF, which I think our other files are also using). So I think it is ok as is, as even editing this file on github would result in these changes.

Rick-Methot-NOAA (approved):
Changes looks good.
Stylistically, should block comments have the /* and */ on separate lines, or OK to have on same line as part of the comment?  I see both approaches being used.
I see some of the block comments getting deleted, which seems OK.  Those blocks probably were copies of other blocks in order to put the variable descriptions in close proximity to code under development.  So, OK to now delete all but one such instance.

k-doering-NOAA:
Thanks @Rick-Methot-NOAA ! Glad you don't see anything that seems harmful. I will work on revising the github action warning run so it takes into account the fewer warnings.

> Stylistically, should block comments have the /* and */ on separate lines, or OK to have on same line as part of the comment? I see both approaches being used.

That's a good question to probably address when we decide on a style guide (#311). I don't have a strong preference either way, as someone who does not use block comments very often!

> I see some of the block comments getting deleted, which seems OK. Those blocks probably were copies of other blocks in order to put the variable descriptions in close proximity to code under development. So, OK to now delete all but one such instance.

Yep, that was intentional on my part. I thought it made the most sense to define the variables once by the function (as you would do with doxygen, for example), rather than have it copy/pasted many times on the function calls, as you pointed out.

Rick-Methot-NOAA (approved):
I approve these changes.

PR #351, opened 2022-07-15
Gha build triggers: reduce runs and standardize syntax

k-doering-NOAA:
## What issue(s) does this PR address? N/A
(note: including the keyword "resolves" means the issue will be closed automatically when merged in.)

## Describe the issue, if not included in the PR

Addresses a request by @Rick-Methot-NOAA to reduce the number of unnecessary runs of gha workflows as much as possible. I found that 2 workflows (running models with estimation, and running models without estimation) could be reduced, by only having them run on the main branch and every pull request to main, rather than running on every push to every branch.

I also made sure workflows had a `workflow_dispatch:` build trigger, which means there is an option to run the workflow manually using a button under the "Actions" tab. 

Finally, I added a one line comment about what each workflow does in the github action workflows.

None, if the github actions pass on this PR, it should be ok to merge.

@Rick-Methot-NOAA should take a look to make sure these changes are as imagined.

No

```

```

Version | Date       | Description               | Issue        | Input change | Action                | Topic                                     | Type  |
| ----- | ---------- | ---------------------- | ------------ | ------------ | --------------------- | ----------------------------------------- | --------------------------------------- |
3.30.20 | [Add date] | [add concise description] | #[add issue] | [yes or no]  | [fix, new, or revise] | [e.g., biology. Use issue label options.] | [input, output, and/or calc, or ALL] |

### DIFF
--- a/.github/workflows/call-build-ss3-warnings.yml
+++ b/.github/workflows/call-build-ss3-warnings.yml
@@ -1,3 +1,5 @@
+# Build stock synthesis with all c++ compiler warnings, and fail job if the
+# number of warnings has increased.
 name:  call-build-ss3-warnings
 on:
   workflow_dispatch:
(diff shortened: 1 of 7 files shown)


k-doering-NOAA:
@Rick-Methot-NOAA , was this one forgotten, perhaps?

PR #370, opened 2022-08-24
add fallthrough statement

k-doering-NOAA:
## Concisely (20 words or less) describe the issues
Removes a warning

## Please Link issue(s)

resolves #[add issue number number], resolves #[optional second issue number if more than 1 issue addressed.]

The call-build-ss3-warnings artifact file should be checked for presence of the implicit fallthrough warning (which should no longer be present):
```
ss.cpp: In function ‘void write_message(int, int)’:
ss.cpp:160:18: warning: this statement may fall through [-Wimplicit-fallthrough=]
  160 |         exitflag = 1;
      |         ~~~~~~~~~^~~
ss.cpp:161:7: note: here
  161 |       case ADJUST:
      |       ^~~~
```

@nschindler-noaa , could you check if the fallthrough is intentional, or if this should be `break;` instead?

-[x] I have documented any new code added (or no new code was added)

```

```

### DIFF
--- a/SS_global.tpl
+++ b/SS_global.tpl
@@ -250,6 +250,7 @@ GLOBALS_SECTION
         break;
       case FATAL:
         exitflag = 1;
+	[[fallthrough]];
       case ADJUST:
       case WARN:
         N_warn++;


k-doering-NOAA:
I checked and the fallthrough statement is indeed gone! The call-build-ss3-warnings is still failing as it is on main because of other warnings (that would need to be fixed on the admb side of things, and may have been fixed already!)

nschindler-noaa (approved):
The fallthrough statement is correct.

PR #374, opened 2022-09-30
Fixes ADMB version number

### LINKED ISSUE #373: Need to update ADMB version info
I was working on updating the SS3 examples in the user-examples repository and I noticed that the version in the ss_new files listed for ADMB is incorrect.
I then found that it is hardwired here
https://github.com/nmfs-stock-synthesis/stock-synthesis/blob/0f70fb715b5ae945946f2ece6e2b1eb62b1fce44/SS_versioninfo_330opt.tpl#L10
We can obviously do a quick fix by just changing the version in the tpl but I think it would be best to somehow get the version of ADMB automatically.
The real question is do we do a fix and then retag the repo with 3.30.0 and put up new executables?

kellijohnson-NOAA:
Updates the version number of ADMB to 13.0, using a hard-coded value.

## Please Link issue(s)
Resolves #373

- [x] I have documented any new code added (or no new code was added)

```

```

### DIFF
--- a/SS_versioninfo_330opt.tpl
+++ b/SS_versioninfo_330opt.tpl
@@ -7,7 +7,7 @@ DATA_SECTION
 !! //  SS_Label_Info_1.1.1  #Create string with version info
 !!version_info += "#V3.30.xx.yy;_fast(opt);_compile_date:_";
 !!version_info += __DATE__;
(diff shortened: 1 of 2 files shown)


Rick-Methot-NOAA (approved):
good

iantaylor-NOAA:
Oops, good catch. I should have done this with the other changes to move to ADMB 13.0. We can do it all again for ADMB 13.1 in the next release.

PR #380, opened 2022-11-17
342 deprecate alk tolerance

### LINKED ISSUE #342: Deprecate ALK tolerance
Discussion on https://github.com/nmfs-fish-tools/SSMSE/issues/129 suggests that the ALK tolerance setting in the starter file should be deprecated. Perhaps we could add a warning first to suggest users set it to 0, then in a future release require users to reset it to 0?

@Rick-Methot-NOAA , feel free to comment on this issue if I didn't capture it how you were imagining

nschindler-noaa:
## Please Link issue(s)

resolves #342

```

```

### DIFF
--- a/SS_ALK.tpl
+++ b/SS_ALK.tpl
@@ -118,7 +118,10 @@ FUNCTION void Make_AgeLength_Key(const int s, const int subseas)
 //                  ALK_range_g_lo(ALK_finder) = column(ALK_range_use, 1);
 //                  ALK_range_g_hi(ALK_finder) = column(ALK_range_use, 2);
 //                }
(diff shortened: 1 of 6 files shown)


Rick-Methot-NOAA (commented):
Thanks Neal.  This looks good.
Since most of code is hardwired to 1, nlength it seems reasonable to do away with ll and lhi and the passing of the vector to calc_ALK.  But definitely OK and workable as you have it now.  What do you think?

nschindler-noaa:
I think it's good as is, but I can clean it up so that range is not even mentioned. Depends on how fast you want this.

Rick-Methot-NOAA (dismissed):

Rick-Methot-NOAA (approved):
Looks good.

PR #398, opened 2023-01-13
fix array bound error in MCEVAL #397

### LINKED ISSUE #397: [Bug]: MCEVAL fatal reporting error with time-varying biology
### Describe the bug

@puntae found a MCEVAL array index error with a model that had time-varying biology.  Tracked it to the save_Gparm which is dimensioned with N rows = N years with forecast.
Overflow occurred because counter for this table was not being reset after MLE, so attempted saves during MCEVAL were beyond the row dimension.  This only occurs when there is time-varying biology because simpler models do not use all the dimensioned rows.

### To Reproduce

per above

### Expected behavior

array index counter gets reset to 0 so that each pass will overwrite this table re-using the same set of rows.

Proposed solution is to move the reset:  save_gparm = 0; to the top of the get_initial_conditions so it will occur in every iteration.  Currently this reset is getting done in other places, which is redundant and not accomplishing the needed reset for MCEVAL.
Note that issue #396 would deprecate the save_gparm approach entirely

### Screenshots

_No response_

_No response_

_No response_

### Additional Context

_No response_

Rick-Methot-NOAA:
move reset of save_gparm such that mceval does not overflow array bound

resolves issue #397 
## Please Link issue(s)

resolves #397 

ran example provided by @puntae
none
yes
- [X] I have documented any new code added (or no new code was added)

```

```

- [X] no further changes to r4ss
- [X] no further changes to the manual
- [X] no further changes to SSI (the SS3 GUI)
- [X] no further changes to the stock synthesis change log (new features, bug reports)

### DIFF
--- a/SS_global.tpl
+++ b/SS_global.tpl
@@ -992,7 +992,6 @@ FINAL_SECTION
     //  SS_Label_Info_12.3.2 #Set save_for_report=1 then call initial_conditions and time_series to get other output quantities
     if (Do_Dyn_Bzero > 0) //  do dynamic Bzero
     {
-      save_gparm = 0;
       fishery_on_off = 0;
       setup_recdevs();
       y = styr;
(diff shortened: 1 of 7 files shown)


Rick-Methot-NOAA:
bypassing review.

PR #404, opened 2023-02-02
remove comperr check rel. Nfleet

### LINKED ISSUE #402: [Bug]: partition weighting not working when ParmSelect > Nfleets
### Describe the bug

I tried to test the partition weighting in the attached model (based on the U.S. west coast 2019 petrale update) where there are 7 fleets and I tried to use D-M parameters to separately weight discard and retained for the first 4 fleets, resulting in 8 parameters.

This setup triggers this message: `Warning 2 Fatal Error! length D-M index for fleet: 4 is: 8 but must be an integer <=Nfleet`
which is triggered by these here:
https://github.com/nmfs-stock-synthesis/stock-synthesis/blob/2257bceea176ed5c648ac65907370a7c451bd5f1/SS_readdata_330.tpl#L1457-L1458
I tried just commenting out those two lines but then get an error
`matrix bound exceeded -- row index too high in imatrix::operator()(int, int) `
which I tracked down to this line:
https://github.com/nmfs-stock-synthesis/stock-synthesis/blob/2257bceea176ed5c648ac65907370a7c451bd5f1/SS_readcontrol_330.tpl#L4018
which I think is because `Comp_Err_Parmloc` is the wrong dimensions due to `Comp_Err_ParmCount` being too small which goes tracks back to the lines that I commented out.

At that point I was just confused, so am handing this off to @Rick-Methot-NOAA.

[2019_petrale_with_partition_weights.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/10552164/2019_petrale_with_partition_weights.zip)

### To Reproduce

Compile current "main" branch or download the artifact from the most recent call-build-ss3 and run with the files in the attached 2019_petrale_with_partition_weights.zip.

### Expected behavior

I expected the model to run with estimates of 8 values for ln(DM_theta) for each combination of discards and retained for the 4 fishing fleets.

### Screenshots

_No response_

_No response_

current "main" branch

### Additional Context

_No response_

Rick-Methot-NOAA:
This check on indexing for comperr is too prone to error and misinterpretation.  It is unnecessary, so is removed from code.

resolves #402 

Code now works with the Petrale example linked to the issue

none 
no need.  code was removed

- [X] I have documented any new code added (or no new code was added)

```

```

- [X] no further changes to r4ss
- [X] no further changes to the manual
- [X] no further changes to SSI (the SS3 GUI)
- [X] no further changes to the stock synthesis change log (new features, bug reports)

### DIFF
--- a/SS_readdata_330.tpl
+++ b/SS_readdata_330.tpl
@@ -1452,12 +1452,7 @@
         min_sample_size_L(parti, f) = 0.001;
       }
   
-      if (Comp_Err_L2(parti, f) > Nfleet)
-      {
-        warnstream << "length D-M index for fleet: " << f << " is: " << Comp_Err_L2(parti, f) << " but must be an integer <=Nfleet ";
-        write_message(FATAL, 1);
-      }
-      else if (Comp_Err_L2(parti, f) > Comp_Err_ParmCount + 1)
+      if (Comp_Err_L2(parti, f) > Comp_Err_ParmCount + 1)
       {
         warnstream << "; length D-M must refer to existing Comp_err definition, or increment by 1:  " << Comp_Err_L2(parti, f);
         write_message(FATAL, 1);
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA (approved):
Looks good and works for the petrale example as you said. 
I'm not sure why my attempted fix didn't work as well, perhaps I got something wrong in the if else combination. Doesn't matter now.

PR #410, opened 2023-02-14
325 make platoon variance ratio a parameter

### LINKED ISSUE #325: make platoon variance ratio a parameter
While working on a project fitting SS3 with 5 platoons to data generated by an IBM, there was evidence that the constant, sd_ratio, that controls the variance of length-at-age within each platoon relative to the variance among the means of the platoons, is not trivial and affects the logL and growth parameter estimates.  Hence it should be estimable.
Currently, sd_ratio is read in ss_readcontrol, and immediately used to calculate sd_between and sd_within.
Then in SS_ALK, sd_between and sd_within are used to calculate the mean and sd of length-at-age for each platoon.

Proposal:
a.  read sd_ratio as a MGparm, rather than a constant.  Seems logical to insert this parameter after the cohort_growth_dev parameter and before the movement parameters.
b.  move calculation of sd_within and sd_between to SS_ALK where they will be estimated dynamically as the parameter, sd_ratio, is estimated.
c.  come up with a better name for the parameter that has the function of sd_ratio.
d.  do this project as the next group coding session to expose group to the indexing of MGparms

nschindler-noaa:
Changes sd_ratio from value to parameter. Adds a parameter to the MG params.

## Please Link issue(s)

resolves #325

[test_platoon_varparm.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/10829744/test_platoon_varparm.zip)

none needed.

Yes.
Modifications made to readcontrol such that an additional MGparm line is read if sd_ratio is < 0.
In ALK, the key change is:
    if (sd_ratio_rd < 0)   {
      dvar_platoon_ratio = MGparm(sd_ratio_param_ptr);

- [x] I have documented any new code added (or no new code was added)

- [x] Yes, there was an input change

in the control file,
```
5 #_N_platoons_Within_GrowthPattern 
-1.09227 #_Platoon_within/between_stdev_ratio (no read if N_platoons=1)
#_Cond sd_ratio_rd < 0: platoon_sd_ratio parameter required after movement params.
 0.031 0.237 0.464 0.237 0.031 #vector_platoon_dist_(-1_in_first_val_gives_normal_approx)
Then
# Movement
# Platoon StDev Ratio 
 0.1 10 1.09227 0.7 99 0 3 0 0 0 0 0 0 0 # Platoon_SD_Ratio
# Age Error from parameters
```

- [ ] no further changes to r4ss

## Describe any changes in r4ss/SS3 manual/SSI that are needed (if not checked):
requires review

### DIFF
--- a/SS_ALK.tpl
+++ b/SS_ALK.tpl
@@ -29,14 +29,14 @@ FUNCTION void Make_AgeLength_Key(const int s, const int subseas)
   {
     ALK_subseas_update(ALK_idx) = 0; //  reset to 0 to indicate update has been done
     gp = 0;
-	if (N_platoon > 1) // calculate the between and within stdev ratio
+	if (N_platoon > 1 && sd_ratio_rd < 0) // calculate the between and within stdev ratio
     {
       plat_sd_ratio = MGparm(sd_ratio_param_ptr);
-      sd_betw_plat = sqrt(1. / (1. + plat_sd_ratio * plat_sd_ratio));
-      sd_with_plat = plat_sd_ratio * sd_betw_plat;
+      sd_betw_plat_parm = sqrt(1. / (1. + plat_sd_ratio * plat_sd_ratio));
+      sd_with_plat_parm = plat_sd_ratio * sd_betw_plat_parm;
 	  platoon_sd_ratio = plat_sd_ratio.xval();
-	  sd_within_platoon = sd_with_plat.xval();
-	  sd_between_platoon = sd_betw_plat.xval();
+	  sd_within_platoon = sd_with_plat_parm.xval();
+	  sd_between_platoon = sd_betw_plat_parm.xval();
     }
     for (int sex = 1; sex <= gender; sex++)
       for (GPat = 1; GPat <= N_GP; GPat++)
@@ -92,8 +92,16 @@ FUNCTION void Make_AgeLength_Key(const int s, const int subseas)
             //  SS_Label_Info_16.3.5  #if platoons being used, calc the stddev between platoons
             if (N_platoon > 1)
             {
-              Sd_Size_between(ALK_idx, g) = Sd_Size_within(ALK_idx, g) * sd_betw_plat;
-              Sd_Size_within(ALK_idx, g) *= sd_with_plat;
+			  if (sd_ratio_rd < 0)
+			  {
+                Sd_Size_between(ALK_idx, g) = Sd_Size_within(ALK_idx, g) * sd_betw_plat_parm;
+                Sd_Size_within(ALK_idx, g) *= sd_with_plat_parm;
+			  }
+			  else
+			  {
+                Sd_Size_between(ALK_idx, g) = Sd_Size_within(ALK_idx, g) * sd_between_platoon;
+                Sd_Size_within(ALK_idx, g) *= sd_within_platoon;
+			  }
             }
 
             if (docheckup == 1)
--- a/SS_ALK.tpl
+++ b/SS_ALK.tpl
@@ -17,8 +17,8 @@ FUNCTION void Make_AgeLength_Key(const int s, const int subseas)
   //  then it calls calc_ALK to make and store the age-length key for that subseason for each biological entity
 
   dvariable plat_sd_ratio;
-  dvariable sd_betw_plat;
-  dvariable sd_with_plat;
+  dvariable sd_betw_plat_parm;
+  dvariable sd_with_plat_parm;
   
   int gstart = 0;
   ALK_idx = (s - 1) * N_subseas + subseas;
--- a/SS_ALK.tpl
+++ b/SS_ALK.tpl
@@ -27,7 +27,7 @@ FUNCTION void Make_AgeLength_Key(const int s, const int subseas)
     gp = 0;
 	if (N_platoon > 1) // calculate the between and within stdev ratio
     {
-      sd_ratio = MGParm(sd_ratio_param_ptr, 3);
+      sd_ratio = value(MGparm(sd_ratio_param_ptr));
       sd_between_platoon = sqrt(1. / (1. + sd_ratio * sd_ratio));
       sd_within_platoon = sd_ratio * sd_between_platoon;
     }
--- a/SS_ALK.tpl
+++ b/SS_ALK.tpl
@@ -35,8 +35,8 @@ FUNCTION void Make_AgeLength_Key(const int s, const int subseas)
       sd_betw_plat = sqrt(1. / (1. + plat_sd_ratio * plat_sd_ratio));
       sd_with_plat = plat_sd_ratio * sd_betw_plat;
 	  platoon_sd_ratio = plat_sd_ratio.xval();
-	  sd_within_platoon = sd_betw_plat.xval();
-	  sd_between_platoon = sd_with_plat.xval();
+	  sd_within_platoon = sd_with_plat.xval();
+	  sd_between_platoon = sd_betw_plat.xval();
     }
     for (int sex = 1; sex <= gender; sex++)
       for (GPat = 1; GPat <= N_GP; GPat++)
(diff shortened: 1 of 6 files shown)


nschindler-noaa:
This needs SSI changes and document changes. Most probably r4ss changes also. the control file has one line deleted (the value) and one line added (the parameter after movement parameters).

Rick-Methot-NOAA (changes requested):
Hi Neal.  Good start on this.
In ALK, do not use value() as it will break the derivative chain and disable estimation of the parameter.
Also, let's consult with Ian regarding whether to just implement this change, or to make it conditional such that existing models that use constant sdratio can still read inputs and work.
@iantaylor-NOAA

nschindler-noaa:
okay, @Rick-Methot-NOAA or @iantaylor-NOAA , what do I use to get the number from the vector? I can't seem to find out anything from the online documentation for ADMB

Rick-Methot-NOAA:
I'll give it a try on one of my examples

Rick-Methot-NOAA:
I think we are close, but gradients are not right.  I tried the derivative checker in ADMB and the numerical derivative was quite different than the analytical.  I will look into this more closely soon.

Rick-Methot-NOAA:
Neal - I think it is working with the changes I added.  Here is my test example. 
[test_platoon_varparm.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/10746550/test_platoon_varparm.zip)

nschindler-noaa:
Looks good to me.

Rick-Methot-NOAA:
I see that the test model check fails when running the "growth_morphs" example.  That is because we did not build in backward compatibility.  I still agree with Ian that we don't need backward compatibility, so let's update the test example so it will use the new version of growth morphs.  @e-gugliotti-NOAA

e-gugliotti-NOAA:
Will update the growth platoons test model after I merge the v3.30.21 test model updates.

Rick-Methot-NOAA:
Hmmm.  We have a sequencing issue because this feature will not be released until 3.30.22.  Perhaps temporarily remove this example from the set until we modify it to use the new approach???

Rick-Methot-NOAA:
@nschindler-noaa 
I have changed my mind regarding backward compatibility.    It seems that it will be fairly straightforward to add for this feature.
Logic goes as:
if Nplatoons > 1 then read sdratio_rd
if sdratio_rd > 0 then copy it to sdratio and use it
else
invoke reading of parameter for sdratio per current code.

iantaylor-NOAA:
Thanks Rick,
This will make the r4ss solution (which we didn't yet embark on) for reading the control file easier more straightforward as well. Sorry I didn't think of something like this earlier.

nschindler-noaa:
I am making some changes for backward compatibility. The control files will need to at least include the sd-ratio value. If there are more that one platoon AND the sd-ratio is negative, then it will look for a parameter.

Rick-Methot-NOAA:
It seems feasible to make reading of sd_ratio_rd conditional on N_platoons>1.

nschindler-noaa:
Yes, like current main. Just adding logic for when sd_ratio_rd is negative.

Rick-Methot-NOAA:
Thanks Neal.  All checks passed so the backward compatibility is working.  I'll take a closer look before approving.

Rick-Methot-NOAA (commented):
I think the conditional logic is a bit more complicated than it needs to be.
In readcontrol, if line 96 changed to sd_ratio_rd = 0
then conditional at line 121 is unnecessary because N_platoon > 1 anytime sd_ratio_rd > 0

It's too bad that we cannot simplify ALK such that it does not need two versions of the code.
Better if we could do:
      if (sd_ratio_rd < 0) plat_sd_ratio = MGparm(sd_ratio_param_ptr);
      else
      plat_sd_ratio = sd_ratio_rd;
      
      but ADMB does not allow setting a dvariable equal to a constant.

kellijohnson-NOAA:
Can't you just trick ADMB and make it a parameter that is not estimated, i.e., with a negative phase?

Rick-Methot-NOAA:
True, and that is how all parameters work when they get assigned an initial value from what is read from the control file.  But the logic of doing that in this case and allowing the backward compatible approach that does not have a parameter seemed complex also.   I would like to avoid extra conditional statements inside of ALK.
I suppose the code could:
a.  create a global  dvariable that is the quantity that will be used in ALK
b.  in the backward compatible case, assign that dvariable to the constant that is read (I think that would happen in PRELIM calcs)
c.  in the new parameter approach, do that assignment dynamicly, probably in the MGSETUP.

nschindler-noaa:
When I create a 'dvariable' before reading the sd ratio, I cannot compile. Even if I change 'number' to 'double'.

Rick-Methot-NOAA:
Current code works and achieves backward compatibility, so no rush.  I can take a look later this week.

nschindler-noaa:
Would it be reasonable to assume that if sd_ratio_rd is negative, then there are more than 1 platoons?

Rick-Methot-NOAA:
sd_ratio_rd is only read if n_platoons is > 1.
suggest assigning it value of 0 if not read

Rick-Methot-NOAA:
The code that I think can be streamlined is in ALK.
This IF statement at line 95 could be eliminated if the value in sd_betw_plat_parm could be set before the "g" loop
			  if (sd_ratio_rd < 0)
			  {
                Sd_Size_between(ALK_idx, g) = Sd_Size_within(ALK_idx, g) * sd_betw_plat_parm;
                Sd_Size_within(ALK_idx, g) *= sd_with_plat_parm;
			  }
			  else
			  {
                Sd_Size_between(ALK_idx, g) = Sd_Size_within(ALK_idx, g) * sd_between_platoon;
                Sd_Size_within(ALK_idx, g) *= sd_within_platoon;
			  }

nschindler-noaa:
sd_ratio_rd is already set to 1. if N_platoons == 1. if sd_ratio_rd is < 0, then it has read a value, therefore there are more than 1 platoons.

nschindler-noaa:
I think the following code will work in SS_ALK:
    if (sd_ratio_rd < 0) // calculate the between and within stdev ratio
    {
      platoon_sd_ratio = MGparm(sd_ratio_param_ptr).xval();
      sd_between_platoon = sqrt(1. / (1. + platoon_sd_ratio * platoon_sd_ratio));
      sd_within_platoon = platoon_sd_ratio * sd_between_platoon;
    }
These values are already computed once when the positive value for sd_ratio_rd is read. Then, the code following is much simpler (and I hope I've got this right this time):
            if (N_platoon > 1)
            {
              Sd_Size_between(ALK_idx, g) = Sd_Size_within(ALK_idx, g) * sd_between_platoon;
              Sd_Size_within(ALK_idx, g) *= sd_within_platoon;
            }

Rick-Methot-NOAA:
I think there needs to be one more change because the final calculation needs to use dvariable, not a constant number.
      instead of:
      platoon_sd_ratio = MGparm(sd_ratio_param_ptr).xval();
 
something like:
     if(sd_ratio_rd>0)
       {dvar_plat_ratio = platoon_sd_ratio;}
  else
      {dvar_plat_ratio  = MGparm(sd_ratio_param_ptr);}

  then create sd_within_platoon and sd_between using dvar_plat_ratio

nschindler-noaa:
Okay.

Rick-Methot-NOAA:
works great and produces a reasonable parameter value with reasonable variance in my test.  ALK code now much more concise.
One addition will be to copy the dvar quantities, which are local to ALK,  to the constant quantities so they can be reported in write_report at line 3804-3806.

minor tweak:  revise write_new text with sdratio_rd so it will reference link to parameter reading

nschindler-noaa:
I'll need to rename the dvar quantities and add a couple lines. 
For the minor tweak, I thought that's what it was doing. sd_ratio_rd is not changed, so the new file would be the same ... oh, use the negative value of the parameter for sd_ratio_rd?

Rick-Methot-NOAA:
No need to rename.  Just set platoon_sd_ratio = value(dvar_platoon_ratio) in ALK.
ditto for the between and within quantities

nschindler-noaa:
Right.

Rick-Methot-NOAA (approved):
Thanks Neal.

Rick-Methot-NOAA:
@chantelwetzel-noaa   @iantaylor-NOAA 
Can you revise the manual to include description of this new feature?
Something like:
If user selects >1 platoon and sets the value of sd_ratio to a negative number, then SS3 will read one additional MGparm line:
#  Movement
#  Platoon StDev Ratio 
 0.1 10 1.09227 0.7 99 0 3 0 0 0 0 0 0 0 # Platoon_SD_Ratio
#  Age Error from parameters
Our experience to date is that this parameter is estimable, especially if there is CAAL data included in the model.   The value of sd_ratio controls the degree to which there is fidelity in the growth trajectories of large length-at-age vs small length-at-age fish.  Our experience is that Linf may be underestimated and degree of dome-selectivity mis-estimated if the platoon approach is not used and the population has high fidelity in growth trajectories.   Parameter values in the range of 0.5 to 1.5 are expected.

PR #436, opened 2023-04-25
fix_compreport_for_sex0

### LINKED ISSUE #435: [Bug]: duplicate entries for unsexed comps getting written to CompReport.sso
### Describe the bug

I just discovered that SS3 is writing the info on unsexed comps to CompReport.sso twice.

I compared across a couple SS3 versions and the issue seems to have appeared in mid-2022, perhaps as a result of this large change: https://github.com/nmfs-stock-synthesis/stock-synthesis/commit/a5ceb77b54ab805b2d0333ffe8e5b57a829748a1. The code to write CompReport.sso is complex and I have not determined where the issue might be, but I'm guessing that some loop intended to write both female and male comps is causing unsexed comps to be written twice.

### To Reproduce

The attached [simple_small_unsexed.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/11314055/simple_small_unsexed.zip) model includes 4 input files plus CompReport.sso. The data file has a single vector of unsexed age comps in 2022 which is shown in CompReport.sso in repeated blocks: lines 1405 to 1419 and again in 1421 to 1435. The "Cum_obs" and "Cum_exp" columns differ because the calculation of those values occurs when the tables are written, but otherwise the entries are identical.

### Expected behavior

No duplicate entries for unsexed data.

### Screenshots

_No response_

_No response_

3.30.20 and 3.30.21

### Additional Context

A while ago I noticed that the shading of the r4ss plots of comps for unsexed fish went away (examples at https://github.com/r4ss/r4ss/issues/785) but couldn't find the cause and it wasn't a big deal so closed the issue.
@chantelwetzel-noaa recently provided a example plot which helped me see that the issue wasn't lack of shading, but the code trying to draw two polygons on top of each other (where the shading was just applied to the space in between them and thus not visible).

I've modified r4ss to find and remove duplicate rows in CompReport (those that are identical in all respects other than "Cum_obs" and "Cum_exp", which I think should never occur under normal circumstances, so there's no rush to get this fix available, but it would be good to include in the next SS3 release.

Rick-Methot-NOAA:
obs with sex=0 were being written twice. Fixed here.
issue #435 

resolves #435

eyeball test
none
not needed

```

```

- [x ] no further changes to r4ss
- [x ] no further changes to the manual
- [x ] no further changes to SSI (the SS3 GUI)
- [x ] no further changes to the stock synthesis change log (new features, bug reports)

### DIFF
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -3963,7 +3963,9 @@ FUNCTION void write_bigoutput()
             }
           }
 
-          for (gg = 1; gg <= gender ; gg ++)
+          int gender2 = gender;
+          if (gen_l(f, i) == 0) gender2 = 1;
+          for (gg = 1; gg <= gender2 ; gg ++)
           {
             if (gen_l(f, i) != 2 && gg == 1)
             {
@@ -4057,6 +4059,7 @@ FUNCTION void write_bigoutput()
           nbins = 0;
           for (gg = 1; gg <= gender ; gg ++)
           {
+            warning<<"obs: "<<i<<" gg "<<gg<<" "<<gen_a(f, i)<<endl;
             if (gen_a(f, i) != 2 && gg == 1)
             {
               z_lo = tails_a(f, i, 1);
@@ -4071,7 +4074,9 @@ FUNCTION void write_bigoutput()
             }
           }
 
-          for (gg = 1; gg <= gender ; gg ++)
+          int gender2 = gender;
+          if (gen_a(f, i) == 0) gender2 = 1;
+          for (gg = 1; gg <= gender2 ; gg ++)
           {
             if (gen_a(f, i) != 2 && gg == 1)
             {
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -4059,7 +4059,6 @@ FUNCTION void write_bigoutput()
           nbins = 0;
           for (gg = 1; gg <= gender ; gg ++)
           {
-            warning<<"obs: "<<i<<" gg "<<gg<<" "<<gen_a(f, i)<<endl;
             if (gen_a(f, i) != 2 && gg == 1)
             {
               z_lo = tails_a(f, i, 1);


Rick-Methot-NOAA:
@e-gugliotti-NOAA  Hi Elizabeth.  Please look at the gha failure for this build.

e-gugliotti-NOAA:
Will do!

e-gugliotti-NOAA:
I'm monitoring the ss3 build action for windows to see if it fails again.

e-gugliotti-NOAA:
The version of R in the github action has updated to 4.3.0 and thus the g++ compiler is a later version. Previous runs that were successful were running with the 4.2.3 R version, also using an earlier g++ compiler. I'm going to try to reverting to an earlier version of R to see if that solves the problem.

iantaylor-NOAA:
That makes sense. Thanks for figuring it out.

Rick-Methot-NOAA:
looks like it is successful with windows build,  then error is with the directory move.  Perhaps missing the \compile in the path?

e-gugliotti-NOAA:
The github action was successfully build here when reverting the version of R to 4.2.3. See successful run [here](https://github.com/nmfs-stock-synthesis/stock-synthesis/actions/runs/4800325600). The big question (at least for me) is why there is always a compiling issue with every R update. I believe the problem is with the g++ compiler and has something to do with the admb command `ad_comm::change_datafile_name`, per the error output [here](https://github.com/nmfs-stock-synthesis/stock-synthesis/actions/runs/4799674396/jobs/8539616855?pr=436#step:20:68). Might be worth asking Johnoel about this potentially.

Rick-Methot-NOAA:
Please go ahead and loop Johnoel in on this question

iantaylor-NOAA:
This is mysterious to me.

Looking back over this issue: https://github.com/admb-project/admb/issues/282, I see that the problem was with gcc 10, but gcc 12 was supported. I also have gcc version 12.2.0 on my computer which matches what is being used in the failing github action, but I'm able to compile SS3 just fine. I will try installing Rtools43 on my computer to see if that causes a problem and then we can reach out to Johnoel for advice if so.

Rick-Methot-NOAA:
good suggestion.  Please do that unless Ian has a strong reason not to

iantaylor-NOAA:
@e-gugliotti-NOAA, that makes sense to me. I'm not aware of any benefit of using to the newer Rtools (which I think we're installing just to get the gcc compiler). We can always circle back later.

e-gugliotti-NOAA:
I've added this topic to the agenda for our next SS3 meeting as it is a reoccurring problem that might have multiple solutions.

e-gugliotti-NOAA (approved):

e-gugliotti-NOAA:
The build ss3 workflow for windows did pass when I manually ran it [here](https://github.com/nmfs-stock-synthesis/stock-synthesis/actions/runs/4812171725). When a workflow is updated, existing pull requests don't use it when you try to re-run it unfortunately but this does pass. @iantaylor-NOAA can you approve your requested change?

iantaylor-NOAA (approved):

iantaylor-NOAA:
Looks good. I should have just approved the first time around.

PR #440, opened 2023-04-26
Fixes #437 regarding partition warning

### LINKED ISSUE #437: [Fix]: Partition warning for lencomp when partition == 2
### Issue description

Lisa, Nathan, and Ian are seeking advice on how to fix an r4ss bug [issue here](https://github.com/r4ss/r4ss/issues/770#issuecomment-1516392063).

### Summary of issue history, discussion, and major aspects of code development

Lisa was able to make the r4ss issue go away by changing the partition in response to this warning:
``` 
Warning 6 : fleet: 1  lencomp has obs with partition==2; will treat as partition=0 because retention not defined; N= 35
```
The model runs fine with the data in either partition 2 or 0, but r4ss crashes due to a mismatch between the "Length_comp_error_controls" output, where fleet 1 is associated only with partition 0, but the "FIT_LEN_COMPS" table and CompReport.sso show the observations associated with partition = 2 as input.

@iantaylor-NOAA said that it would be nice to avoid having r4ss crash in situations like this in the future, a few options to do this include:
- change r4ss so to treat part = 2 as part = 0 if there is no retention for this fleet
- change SS3 to write part = 2 in Length_comp_error_controls
- changing SS3 to write part = 0 in all output rather than preserving the input value of part = 2 in most places
- [more extreme] changing SS3 to stop instead of continuing when you enter data as part = 2 without having retention defined for that fleet

Nathan suggested:
```
Warning 6 : fleet: 1  lencomp has obs with partition==2; will treat as partition=0 because retention not defined, this change should also be made by the user in the original data file to avoid errors; N= 35
```

@Rick-Methot-NOAA noted that the warning message is created in ss_prelim, which is after both data and control have been read.  That gives us easier options for a fix.  We could just expand the warning as you suggest, or we could have SS3 revise the internal value for partition to 0 if do_retain=0 for that fleet.  That would be Rick's preferred solution.

A test case is available [here](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/11502655/partition_test.zip)
### Items to do

- [x] @Rick-Methot-NOAA to revise the internal SS3 value for partition to 0 if do_retain=0 for the fleet with where partition==2. 
- [ ]  Need to revise writing of data_echo, data_expval and data_boot so the original read value is output for all of lencomp, agecomp and meansize comp.  @nschindler-noaa Can you look into this needed change?

### Does documentation already exist in the SS3 User Manual?

- [X] Yes, the link to the section in the SS3 User Manual is provided below.
- [ ] Yes, but further documentation needs to be added and is provided below.
- [ ] No, the documentation that should be added to the SS3 User Manual pertaining to this issue is provided below.

### Documentation to add to the SS3 User Manual OR link to existing documentation

- Documentation can be found in 7.14 section of the User Manual linked [here](https://nmfs-stock-synthesis.github.io/doc/SS330_User_Manual.html#length-composition-data-structure)

### Are r4ss changes needed?

- [x] No, this issue doesn't require changes to r4ss
- [ ] Yes, this issue requires changes to r4ss (if checking this box, please add iantaylor-NOAA as an Assignee in the panel to the right).
- [ ] I don't know and would like an r4ss consult (if checking this box, please add iantaylor-NOAA as an Assignee in the panel to the right).

### Are SSI changes needed?

- [X] No, this issue doesn't require changes to SSI
- [ ] Yes, this issue requires changes to SSI (if checking this box, please add nschindler-noaa as an Assignee in the panel to the right).
- [ ] I don't know and would like an SSI consult (if checking this box, please add nschindler-noaa as an Assignee in the panel to the right).

Rick-Methot-NOAA:
revision of output when compdata read with 2 as retained, but retention not defined
## Please Link issue(s)
#437
resolves #437 
eyeball test 
code is not consistent regarding ssnew files.  Data.echo OK, but for other new files need to decide if the original or transformed partition code should be output.  lencomp writes out the revised value of mkt_l(f,i), but agecomp and meansize write out saved value of header_a which is a vector containing the original value of mkt_a, etc.
yes

```

```

- [ ] no further changes to r4ss

### DIFF
--- a/SS_prelim.tpl
+++ b/SS_prelim.tpl
@@ -1131,6 +1131,7 @@
       for (i = 1; i <= Nobs_l(f); i++)
       {
         parti_cnt(abs(mkt_l(f, i)))++;
+        if (Do_Retain(f) == 0) mkt_l(f,i) = 0;  //  force to partition 0 if retention not defined
       }
       if (parti_cnt(1) > 0 && Do_Retain(f) == 0)
       {
@@ -1139,7 +1140,7 @@
       }
       if (parti_cnt(2) > 0 && Do_Retain(f) == 0)
       {
-        warnstream <<  "fleet: " << f << "  lencomp has obs with partition==2; will treat as partition=0 because retention not defined; N= " << parti_cnt(2);
+        warnstream <<  "fleet: " << f << " lencomp has N obs with partition==2 (retained); changed to partition=0 because retention not defined; N= " << parti_cnt(2);
         write_message (WARN, 0);
       }
       if (parti_cnt(2) > 0 && (fleet_type(f) == 2 || seltype(f, 2) == 3 || seltype(Nfleet + f, 2) == 3)) //  error if retained catch obs are with no retention fleets
@@ -1155,6 +1156,7 @@
       for (i = 1; i <= Nobs_a(f); i++)
       {
         parti_cnt(abs(mkt_a(f, i)))++;
+        if (Do_Retain(f) == 0) mkt_a(f,i) = 0;  //  force to partition 0 if retention not defined
       }
       if (parti_cnt(1) > 0 && Do_Retain(f) == 0)
       {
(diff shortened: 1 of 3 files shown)


e-gugliotti-NOAA:
@Rick-Methot-NOAA this is another situation where the github action doesn't run the updated github action but build ss3 for windows does pass when run manually as seen [here](https://github.com/nmfs-stock-synthesis/stock-synthesis/actions/runs/4812317432).

e-gugliotti-NOAA (approved):
Ran updated ss3 with Rick's example and compared with previous ss3; looks like the Part has indeed changed to 0 instead of 2 in the Report.sso and a warning ("fleet: 1 lencomp has N obs with partition==2 (retained); changed to partition=0 because retention not defined; N= 35") is produced in the warning.sso. Looks good to me.

Issue #416, opened 2023-03-02
[Refactor]: combine workflows repo with stock-synthesis?


iantaylor-NOAA:
### Refactor request

@k-doering-NOAA and I were discussing how to keep the github actions working well with the least effort and concluded that it would be a good idea to move all the files from the `workflows` repo into the `stock-synthesis` repo. 

In the time since the `workflows` repo was first created, github actions have evolved and we have learned a lot and also discovered some of the challenges of having them in a separate place, such as the need to create a separate branch of `stock-synthesis` to test any branch in `workflows` when changing the github actions, rather than a single branch of a combined repository. 

Some of the files in the `workflows` repo are used by `doc`, rather than `stock-synthesis` and those could either be moved into the `doc` repo directly or we could just update which repo is referenced in lines like 
```
uses: nmfs-stock-synthesis/workflows/.github/workflows/build-ss3-manual-html.yml@main
```

I think the primary drawback of this approach is that changes to the github actions will now generate notifications for the watchers of `stock-synthesis` who are more numerous than the few watchers of `workflows`. However, now that these github actions are working well there are many fewer notifications than there were when they were first getting set up.

I'm happy to do the refactoring, perhaps with help with @k-doering-NOAA if needed, but want to post this as an issue to see if anyone has a problem with the proposed change or suggestions on whether to put all the files into `stock-synthesis` or distribute them to the package that uses them (I think `doc` may be the only other one).

### Expected behavior

Fewer moving parts to maintain and keep track of.

### Progress on Issue
|workflows|to|complete|
|---|---|---|
build-ss3-manual-html.yml | doc | :x:; PR in review |
build-ss3-manual-pdf.yml | doc | :x:; PR in review |
build-ss3-warnings.yml | stock-synthesis | ✔️ |
build-ss3.yml | stock-synthesis | ✔️ |
check-ss3-models-run.yml | user-examples | ✔️ |
deploy-ss3-docs.yml | doc | :x:; PR in review |
r4ss-extra-tests.yml | r4ss | ✔️ |
run-ss3-mcmc.yml | stock-synthesis | ✔️ |
run-ss3-no-est.yml | stock-synthesis | ✔️ |
run-ss3-with-est.yml | stock-synthesis | ✔️ |
test-r4ss-with-ss3.yml | stock-synthesis | ✔️ |
update-ss3-version.yml | ss3sim | :x:; PR in review |
update-ss3-version.yml | ssmse | :x: |

[...]

PR #448, opened 2023-05-25
Transfer workflows from workflow repo

e-gugliotti-NOAA:

Transfer workflows from the workflows repo to the stock-synthesis repo. 

* Part of issue #416 

Successful test GitHub action runs on the transfer-workflows branch are linked in the issue and [here](https://github.com/nmfs-stock-synthesis/stock-synthesis/actions?query=branch%3Atransfer-workflows)


iantaylor-NOAA (approved):
Everything looks good to me. Thank you @e-gugliotti-NOAA.

Among other benefits of this changes is that when looking at the results of a github action, it's nice to be able to click on the "Workflow file" link in the lower left and have it provide the actual workflow and not just the "call_..." wrapper that points to the other repo.

kellijohnson-NOAA (approved):
Thanks @e-gugliotti-NOAA for making all the reusable workflows "normal" workflows in this PR. I minimally checked the changes to the files, largely just to make sure that each workflow was a useable rather than "reusable" workflow. In the future, I think reviewing this would have been easier if you would have transferred the file with all of the history rather than just copying and pasting the file contents as a new file into this repository. That way it would have been very clear what changed from the old reusable workflow to this "new" file. But, what you did is 100% sufficient and can be merged into main in my opinion. 

I am not sure if we should squash the commits into one. That is maybe a question for @Rick-Methot-NOAA if you would like to keep the commits that test this within the feature branch or have all commits act as a single commit where you just see the end product.

Issue #466, opened 2023-06-06
allow late recdevs to be expected value for an index


Rick-Methot-NOAA:
### Describe the solution you would like.

Currently the use of recdevs as the expected value for a survey index is limited by recdev_end:
                  case 36: // recruitment deviation  #36
                  {
                    if (y >= recdev_start && y <= recdev_end)
                    {
                      vbio = recdev(y);
                    }

This constraint is not necessary and prevents use of an index in the late period and forecast.  

### Describe alternatives you have considered

none seem feasible

### Statistical validity, if applicable

Removing constrain will enable forecasts to be informed by env indexes

### Describe if this is needed for a management application

petrale assessment would have used this

Rick-Methot-NOAA:
@iantaylor-NOAA @kellijohnson-NOAA @chantelwetzel-noaa   FYI

also, working this issue reminds me of need to prioritize #5  so the mean offset can become an estimated parameter

e-gugliotti-NOAA:
@Rick-Methot-NOAA, in the manual the special surveys only go up to 35 (see link to indices section of manual [here](https://nmfs-stock-synthesis.github.io/doc/SS330_User_Manual_release.html#indices)), could you share with me what text to include for special survey 36?

e-gugliotti-NOAA:
Okay great, option 31 already has
`31 = exp(recruitment deviation), useful for environmental index affecting recruitment`,
it was just 36 that wasn't defined in the documentation yet.

PR #474, opened 2023-07-11
implement changes and warnings for recr surveys

Rick-Methot-NOAA:
see info with the issue

* Resolves issue #466

## What tests have been done? 
temporary edits to simple example were used to test display of the new warnings
[recr_svy_test.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/12017818/recr_svy_test.zip)

### DIFF
--- a/SS_expval.tpl
+++ b/SS_expval.tpl
@@ -334,6 +334,8 @@ FUNCTION void Get_expected_values(const int y, const int t);
                   }
                   case 31: // exp(recruitment deviation)  #31
                   {
+                    vbio = mfexp(recdev(y));
+                    /* 
                     if (y >= recdev_start && y <= recdev_end)
                     {
                       vbio = mfexp(recdev(y));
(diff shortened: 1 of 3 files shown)


e-gugliotti-NOAA:
I ran the growth_timevary test model which has a survey where the unit is 32 and the endyear of that survey is 2009, but the last main recdev year is 2008. I get the following:
`Note 2 Suggestion: Recruitment survey: 4 has data in: 2009, which is after last main recdev: 2008. Suggest extend recdev_end, or change fore_devs phase: 6`
which looks like it's doing what it's supposed to given the code you added.

Rick-Methot-NOAA:
The example you tried had a forecast phase = 6, which is good and so doesn't need change.  I will add a condition such that that warning will only appear if forecast phase is <=0.

e-gugliotti-NOAA:
I ran the test model that you attached and got the note in the warning.sso. I also ran the growth_timevary model which has a survey where the unit is 32 and the last year of that survey is 2009 but the last year of main recdev is 2009. I get the following note:
`Note 2 Suggestion: Recruitment survey: 4 has data in: 2009, which is after last main recdev: 2008. Suggest extend recdev_end, or change fore_devs phase: 6`
so it seems things are working as they should.

e-gugliotti-NOAA (approved):

Rick-Methot-NOAA:
Thanks.  Please edit the manual per the text I put in the issue #466

e-gugliotti-NOAA (approved):

PR #491, opened 2023-08-01
repair_check_on_Nparms_in_parfile

Rick-Methot-NOAA:
The checksum value at end of ss.par was intended to provide a check when the number of parameters in the model changed.  This check is now added to ss_prelim.  SS3 will exit if value of checksum, which is not actually a checksum, is not equal to 999.

No issue is linked
No input change for the user.  The 999 value is created when a model is read from the control file.


e-gugliotti-NOAA (approved):
Changed number at end of ss.par and got that fatal warning.

e-gugliotti-NOAA (approved):

Issue #499, opened 2023-08-30
[Feature]: add output of vulnerable biomass and max_sel for each fleet


Rick-Methot-NOAA:
### Describe the solution you would like.

While working with a user regarding SS3 output of vulnerable biomass for fisheries (vuln_bio in Report 22 for Survey CPUE), I realized that the output of selected (vulnerable) biomass is not what users expect for fisheries.

If the index is a survey, then SS3 uses Z to get biomass at a particular time and report it as vuln_bio:
                tempvec_a = elem_prod(natage(t, p, g), elem_prod(mfexp(-Z_rate(t, p, g) * timing), sel_a(y, f, gg)));
but if a fishery, then it uses:
                  tempvec_a = elem_prod(natage(t, p, g), elem_prod(Zrate2(p, g), sel_a(y, f, gg)));
which if multiplied by F would get the catch
where Zrate2 is from the baranov catch equation:
                    Zrate2(p, g) = elem_div((1. - mfexp(-seasdur(s) * Z_rate(t, p, g))), Z_rate(t, p, g));

SOLUTION:

- [x] create two new arrays for output:  vuln_bio(t,f) and vuln_num(t,f)
populate these arrays with output only during the "bigreport" stage.
```
              vuln_bio(t, f) += sel_bio(s, f, g) * elem_prod(natage(t, p, g), mfexp(-Z_rate(t, p, g) * 0.5 * seasdur(s)));
              vuln_num(t, f) += sel_num(s, f, g) * elem_prod(natage(t, p, g), mfexp(-Z_rate(t, p, g) * 0.5 * seasdur(s)));

```
Note that it uses middle of season because selectivity and bodywt are referenced to middle of season.
By use of middle of season, it is not equivalent to summary biomass which is calculated at beginning of season.
Also, users should note that selectivity may be < 1.0 for all ages.
Output occurs as 2 columns in CATCH table of report.sso
```
CATCH report:15
# where vuln_ is mid-season selected bio or numbers; sel_ is selected total catch; dead_ is catch without live discards; ret_ is retained catch
Fleet Fleet_Name Area Yr Seas Time Obs Exp Mult Exp*Mult se F  Like vuln_bio sel_bio dead_bio ret_bio vuln_num sel_num dead_num ret_num
```
- [x] provide report of maximum ASEL2 selectivity.  Occurs immediately after the AGE_SELEX report:32 table
```
maximum ASEL2
Fleet fleet_name year seas max
[...]

PR #500, opened 2023-08-31
add vuln_bio and max_sel to report

Rick-Methot-NOAA:
addresses issue #499 

## What tests have been done? 
eyeball test.  Example output included in issue #499 

None

### DIFF
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -1305,7 +1305,7 @@ FUNCTION void write_bigoutput()
   {
     SS2out << endl
            << pick_report_name(15) << endl;
-    SS2out << "Fleet Fleet_Name Area Yr Seas Time Obs Exp Mult Exp*Mult se F  Like sel_bio kill_bio ret_bio sel_num kill_num ret_num" << endl;
+    SS2out << "Fleet Fleet_Name Area Yr Seas Time Obs Exp Mult Exp*Mult se F  Like sel_bio kill_bio ret_bio sel_num kill_num ret_num vuln_bio_mid vuln_num_mid" << endl;
     for (f = 1; f <= Nfleet; f++)
     {
       if (fleet_type(f) <= 2)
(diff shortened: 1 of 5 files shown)


e-gugliotti-NOAA:
Fixing workflows to also run when PR goes from draft to ready for review, apparently that also needs to be specified.

e-gugliotti-NOAA:
Rick you might need to switch this back to a draft and then back to ready for review again for the actions to run now that I pushed the fix.

Rick-Methot-NOAA:
no need to full set of tests as this is only reporting.

e-gugliotti-NOAA (dismissed):
Tested locally and getting the new output in report 15 (vuln_bio_mid & vuln_num_mid) and maximum ASEL2.

kellijohnson-NOAA:
@Rick-Methot-NOAA I am on vacation for the holiday weekend so I will not be able to look at this until next week.

e-gugliotti-NOAA (approved):
Changes look good and are working locally for me.

SS_write_report.tpl
```diff
@@ -1305,7 +1305,7 @@ FUNCTION void write_bigoutput()
   {
     SS2out << endl
            << pick_report_name(15) << endl;
-    SS2out << "Fleet Fleet_Name Area Yr Seas Time Obs Exp Mult Exp*Mult se F  Like sel_bio kill_bio ret_bio sel_num kill_num ret_num" << endl;
+    SS2out << "Fleet Fleet_Name Area Yr Seas Time Obs Exp Mult Exp*Mult se F  Like sel_bio kill_bio ret_bio sel_num kill_num ret_num vuln_bio_mid vuln_num_mid" << endl;
```
kellijohnson-NOAA, line 1308: Do we care that there is now a mix of biomass and numbers rather than before all biomass columns came before the numbers columns, either can be accommodated in r4ss, I am just trying to think of others? Maybe it is best to tack them on the for those that use numerical reference to columns but for aesthetic and future code it seems more logical to keep them grouped. Second, I would have named the new columns `vuln_mid_bio` and `vuln_mid_num` to retain the _bio and _num at the end of the column name rather than placing it in the middle.

Rick-Methot-NOAA: kill and ret don't apply to the vuln calculation.
all other columns use "mid" also, so perhaps just drop that qualifier?
so could be:  vuln_bio sel_bio kill_bio ret_bio vuln_num .... 
then perhaps but the defining eqs above or below the table?

Issue #505, opened 2023-09-22
[Bug]: check length-at-age output when settlement is at age 1.


Rick-Methot-NOAA:
### Describe the bug

unexpected kink in length-at-age output when settlement is at age 1 was observed in a model at the ICES course.

### Expected behavior

smooth curve

### Which version of SS3 are you seeing the problem on?

3.30.21

### Additional Context

I could not find a problem when I checked this out using petrale model which uses Recr at age 0 and Amin = 1.  Compared to a revised model that had recr at age 1 and amin at age 0 (which is relative to the age at recruitment.  Calculated mean size-at-age matches at age 1 and above.  Below age 1 they differ because one of the models has no fish below age 1 so assigns a mean size for those ages to be same as size at recruitment.
![image](https://github.com/nmfs-stock-synthesis/stock-synthesis/assets/6250016/3cb33603-1168-467a-845f-c1b1b41276ed)

Despite the comparable results, I think we should steer the community towards:  spawn in year Y then recruit at age 1 in year Y+1.  That is a much more common practice in other models.  The SS3 usage of recruitment at age 0 was largely driven by the need to accommodate the Santa Cruz juvenile (age 0) rockfish survey.  Then it got enshrined in the "simple" example.  Your thoughts on this appreciated  @nmfs-stock-synthesis/developers

Rick-Methot-NOAA:
I did see one anomaly regarding calc of Linf when A2 is 17 rather than 999 and recr is at age 1.  Will followup on that.

e-gugliotti-NOAA:
Right now the manual has the following:
```
1 & & Growth Amin (A1): Reference age for first size-at-age L1 (post-settlement) parameter. 
25 & & Growth Amax (A2): Reference age for second size-at-age parameter L2 (999 to use as L infinity).
```
Are you wanting the wording to be _exactly_ how it is in the revised description that you put above? Something like this:

```
1 & & Age (post-settlement) for L1 (a.k.a. Amin). First growth parameter is the size at this age; linear growth below this.
25 & & Age (post-settlement) for L2 (a.k.a. Amax). Use 999 to treat as L infinity (Linf).
```

Rick-Methot-NOAA:
The important change to make is to add "post-settlement" to the Amax line.

PR #509, opened 2023-09-28
modify ssnew text for growth Amax

Rick-Methot-NOAA:
* Resolves issue #505 

## What tests have been done? 
careful comparison of echoinput and mean_size table for alternative values of recruitment age

none


e-gugliotti-NOAA (approved):
Looks good to me and doc is updated already

PR #521, opened 2023-10-20
fix MSY search output with >1 season

Rick-Methot-NOAA:
format of the MSY search output was corrupted when there is >1 season.  Fix by using colsum to output the sum of catch across seasons for each fleet.

There is no issue.

## What tests have been done? 
examined revised output and verified that it was correct to use colsum, not rowsum

[x] No test files are required for this pull request.


e-gugliotti-NOAA:
@Rick-Methot-NOAA, can you walk me through where these changes are showing up in the report file a bit more? At first glance, the logic seems fine but I'm having a hard time finding where it will show up. @iantaylor-NOAA Maybe you could help me out with this since Rick is at PICES this week?

Rick-Methot-NOAA:
The output is only in forecast-report.sso and is not used by any R read
functions that I know of.  Basically it is just for my investigation of
user problems when they occur.

e-gugliotti-NOAA (approved):

PR #533, opened 2023-12-05
Parallelize workflows

e-gugliotti-NOAA:

The following workflows have been parallelized using the {furrr} package:

- run-ss3-with-est
- test-r4ss-with-ss3
- run-ss3-no-est (via the ss3-test-models updated R script run_from_par.R)

See successful workflows on the branch [here](https://github.com/nmfs-ost/ss3-source-code/actions?query=branch%3Aparallel-est-workflow); these were run by changing the workflow trigger to run workflows on all pushes (this was changed back to the original setting for the PR)

None.

### DIFF
--- a/.github/workflows/run-ss3-no-est.yml
+++ b/.github/workflows/run-ss3-no-est.yml
@@ -4,8 +4,8 @@ name: run-ss3-no-est
 on:
   workflow_dispatch:
   push:
-    # paths:
-    #   - '**.tpl'
+    paths:
+      - '**.tpl'
     # branches: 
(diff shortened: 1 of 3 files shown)


iantaylor-NOAA (approved):
Looks good.

It looks like run-ss3-with-est is the slowest workflow and parallelizing speeds it up from about 29 minutes to 17 minutes, which will make a big difference.

One thing worth experimenting with in the future might be removing the `omit = 1` entirely to see if that speeds things up further since the omitted core might not have much other work to do in a github action.

Issue #452, opened 2022-08-24
improve build_ss3_warnings github actions


k-doering-NOAA:
I added a [reference compiler warnings file](https://github.com/nmfs-stock-synthesis/workflows/blob/main/reference_files/warnings_ss_ref.txt) to the workflows repo in order to facilitate checking failing build_ss3_warnings jobs. Note that there are 2 new warnings on the main branch relative to this older file, so we may want to update this file with what we accept as a new, acceptable reference file.

Right now, the job just counts the number of warnings and see if it is greater than the expected number. Then, there is a manual step for the user to figure out which warnings are different. It would be better if gha could also highlight which of the warnings are new compared to a reference file.

PR #536, opened 2023-12-13
Warning action update

e-gugliotti-NOAA:

Update build-ss3-warnings.yml to compare to a reference warnings file and fail if the number of warnings from a new build is greater than the number of warnings in that reference warnings file.

* Resolves issue #452 

None.


Rick-Methot-NOAA (dismissed):
looks good

iantaylor-NOAA:
This is great, thank you @e-gugliotti-NOAA.

Is it worth making a plan for when to update the .github/workflows/reference_files/warnings_ss_ref.txt file?
I can think of at least 2 scenarios:
* If the github action fails due to an increase in length of the warnings files, and we can't make the new warning go away but decide it can be ignored, then we would want to update the reference file to account for the increased length. 
* Any time a warning goes away we would also want to update the file to make sure future new warnings get detected.
 
Would it be worth supplementing the `if (n_warn > length(ref))` to with a separate failing message when ``if (n_warn < length(ref))` to help us see when the second scenario occurs? Maybe that's not needed since the change that makes the warning go away is probably some change that we're making to achieve that end and won't just happen by accident.

e-gugliotti-NOAA:
@iantaylor-NOAA I added lines similar to what you suggested. The R code written by Kathryn originally to get the n_warn only detects true warnings, so it isn't detecting the 2 lines that say "In file included from ss.cpp:7:" currently in the reference warnings text file, which is why the code has `if(n_warn+length(ref) > length(ref))`. Could you add your review?

iantaylor-NOAA (approved):
Good catch with the 2 lines that we're ignoring.

Issue #534, opened 2023-12-07
[Bug]: Model run intermittently exits while writing bootstrap data file


iantaylor-NOAA:
### Describe the bug

I've recently discovered multiple models (including multiple variations of simple_small in r4ss and the 2021 spiny dogfish assessment) exiting the run early when using 3.30.22. The echoinput file ends with `Begin writing bootstrap data file(s)` instead of continuing on to `Write starter.ss_new file`.

I don't think this is related to the bootstrap data itself for two reasons:
1. The most recent case exited just prior to writing `1 #_spawn_month` in data_boot_001.ss (see screenshot below), as opposed to exiting when while writing randomly generated data. There's nothing in between these two lines where it exited that could explain the exit: https://github.com/nmfs-ost/ss3-source-code/blob/797760ff1ae10554a4f7729b643ab275125981c0/SS_write_ssnew.tpl#L81-L82
2. Running the model again with the random seed from the model with incomplete data file produced a complete data_boot_001.ss 

### To Reproduce

Running the attached files with the 3.30.22 executable on Windows sometimes reproduces the error and sometimes doesn't.
[simple_small_biasadjQ_varySE.zip](https://github.com/nmfs-ost/ss3-source-code/files/13606223/simple_small_biasadjQ_varySE.zip)

I ran using the command `ss3 -nohess -stopph 0` again and again. The first 5 quit after 
```
Begin writing *.ss_new output files ...
```
and the 6th time it completed:
```
Begin writing *.ss_new output files ... Finished writing *.ss_new output files

!!  Run has completed  !!
!!  See warning.sso for 2 notes

Finished running model 'ss3' after 0.68 s.
```

### Expected behavior

Finish writing bootstrap data files and remaining .ss_new files.

### Screenshots

Screenshot shows matching random seeds in the completed and incomplete bootstrap data files.
![image](https://github.com/nmfs-ost/ss3-source-code/assets/4992918/bfca6dcf-11e6-42c5-a756-633f714227e3)

### Which OS are you seeing the problem on?

Windows

### Which version of SS3 are you seeing the problem on?

3.30.22

Rick-Methot-NOAA:
Totally bizarre.  I get 5-6 good runs, then a failure, then repeat that cycle.

Changed random seed to  1677620645
and now get complete boot data and failure in dynamic Bzero.

Change seed again and get stop in another location.

Hypothesis is that we have an uninitialized variable

[...]

PR #537, opened 2023-12-13
#534 repair string error in bootstrap data header

Rick-Methot-NOAA:
Resolves issue #534 

## What tests have been done? 
Before this change, the program would exit at or after writing the header info for the first bootstrap file.  This would occur about every 5 runs.  After the change, I encountered no exiting even after >10 runs.

- [ ] verify that r4ss did not read the line that now simply says "bootstrap"

### DIFF
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -1355,6 +1355,8 @@ FUNCTION void write_bigoutput()
     }
   }
   int bio_t;
+  dvector Bio_Comp(1, N_GP * gender);
+  dvector Num_Comp(1, N_GP * gender);
   // REPORT_KEYWORD 16 TIME_SERIES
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA (approved):
Looks good. I don't understand why the string error was causing the problem, but great that you fixed it. 

Removing the number in the lines like `#_bootdata:_3` is no loss since line 74 of SS_write_ssnew.tpl is providing more detailed output like `#_bootstrap file: 1  irand_seed: 1701981559 first rand#: 0.864036`.

Since the use of the bootstrap data files is relatively rare, and this may not occur for all models generating the files, do you think this can wait for the next release to be available to SS3 users or is a fix release needed?

Rick-Methot-NOAA:
I think hake will need this fix.

PR #542, opened 2023-12-19
restore sizecomp bootstrap output

Rick-Methot-NOAA:
resolves issue #541 

uses code that was created in issue #357  but accidentally deleted

Testing can use this zip, which may be a good candidate for adding to the test repo
[D-M_test.zip](https://github.com/nmfs-ost/ss3-source-code/files/13709772/D-M_test.zip)

Testing demonstrates that data.expval and data.boot1 are different from data_echo.

Issue #541, opened 2023-12-18
[Bug]: Size Frequency section in data_boot_001.ss and data_expval.ss identical to input file


e-gugliotti-NOAA:
### Describe the bug

While testing PR #540 to fix the size frequency method in CompReport.sso, @iantaylor-NOAA saw that the size frequency section in data_boot_001.ss and data_expval.ss are identical to the input file (but were different in 3.30.19).

It looks like this commit https://github.com/nmfs-ost/ss3-source-code/commit/61d2cba7ea9ce1d640920ec242c54098f039443e, which included a variety of other changes, may have inadvertently led to the observed output getting written to all three data file types.

In particular this change looks incorrect: https://github.com/nmfs-ost/ss3-source-code/commit/61d2cba7ea9ce1d640920ec242c54098f039443e#diff-17a83800a032b2061d3ddd5c6da9d766d9796ecb81f2ebf1599b9808c206247eL775-R782
but the change includes moving from looping over size frequency observations to size frequency methods, so I'm assuming the fix is not as simple as swapping SzFreq_obs and SzFreq_exp.

Lesson from all this: more tests related to generalized size frequency and/or bootstrap data would be helpful.
The one test model with generalized size frequency data has the same scale and units for both size frequency methods: https://github.com/nmfs-ost/ss3-test-models/blob/main/models/two_morph_seas_areas/data.ss#L455. And I don't think we have any tests for expected values or bootstrap data.

### To Reproduce

See PR #540.

### Expected behavior

Size frequency section in data_boot_001.ss and data_expval.ss should be different from the input file.

### Which version of SS3 are you seeing the problem on?

Versions after v.3.30.19

iantaylor-NOAA:
@Rick-Methot-NOAA, should we consider a bug fix release once this issue is resolved?
Yohei Tsukahara is interested in bootstrapping for Pacific bluefin tuna.
I don't know if there are other issues that could be rolled in as well.

e-gugliotti-NOAA:
Just created a 3.30.22.1 milestone

Rick-Methot-NOAA:
I looked at the code and see how an errant cut 'n paste involving:
https://vscode.dev/github/nmfs-ost/ss3-source-code/blob/main/SS_write_ssnew.tpl#L398-L401
created this problem.  Should be straightforward to restore the expval and boot code.

PR #543, opened 2023-12-19
use different artifact names

e-gugliotti-NOAA:

For some reason, the build-artifact workflow is no longer uploading all of the executables from the different operating systems to the same artifact named "ss3" (see artifact from PR #542). I have changed the workflow to upload an artifact for each OS, which is working (see workflow run [here](https://github.com/nmfs-ost/ss3-source-code/actions/runs/7263287295)).

First seen in PR #541 by @iantaylor-NOAA.

None. 

### DIFF
--- a/.github/workflows/build-ss3.yml
+++ b/.github/workflows/build-ss3.yml
@@ -267,5 +267,5 @@ jobs:
         if: success()
         uses: actions/upload-artifact@main
         with:
-          name: 'ss3'
+          name: ss3-${{ matrix.config.os }}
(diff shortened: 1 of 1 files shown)


Rick-Methot-NOAA (approved):
hope this works.  Is it the reason my gha is failing?

e-gugliotti-NOAA:
I think so, I don't know why the upload-artifacts action now doesn't want to upload artifacts in a folder under the same name anymore but this should allow all artifacts to upload now.

Issue #535, opened 2023-12-13
[Feature]: implement survey errtype that adds lognormal bias adjustment


Rick-Methot-NOAA:
### Describe the bug

Expected to see different survey Q when bias_adj flag is turned on

### To Reproduce

Examine the survey INDEX01 output for a survey that does and a survey that doe not have different se among the obs.  Do this with four conditions:  bias_adj on vs. off; float on vs. off.
use this example:
[simple_small_biasadjQ_varySE.zip](https://github.com/nmfs-ost/ss3-source-code/files/13664275/simple_small_biasadjQ_varySE.zip)

### Expected behavior

different Q

### ToDo

- [ ] create new options for svy_errtype:  -2=gamma (Cole); -1=normal; 0=lognormal; >1XXX=T.  Add 1 = lognormal with bias adj..  Note that this means that the T-dist cases must be changed to use the revised code.
- [ ]  Explain to user the use of  bias adjustment inherent to error_type as input in data file, vs. explicit bias_adjustment flag in control file.

Rick-Methot-NOAA:
fleet 1 has varying se; fleet 2 has same se each year
--- bias_adj  float  fleet1_q   fleet2_q
0              0

e-gugliotti-NOAA:
@Rick-Methot-NOAA Would it be worth including a warning when people turn on bias adjust but don't also turn on float, to go back and do that?

Rick-Methot-NOAA:
Yes!  That makes sense.  Will do.

Rick-Methot-NOAA:
Larry's work in 2012 also added lognormal with bias adjustment with intention of using it in the prior on Q:
https://vscode.dev/github/nmfs-ost/ss3-source-code/blob/main/SS_objfunc.tpl#L1587-L1597

Rick-Methot-NOAA:
[Qmethod2.xlsx](https://github.com/nmfs-ost/ss3-source-code/files/13679526/Qmethod2.xlsx)

Rick-Methot-NOAA:
Here is my current state of some progress:

Two surveys, one with constant observation se and the other with time-varying se
Four scenarios: Float; float with bias corr; parm; parm with bias corr
Note that bias corr with float is invoked in the control file; but bias adj with params is invoked with err_type in the data file!!!  Who designed this?

Results:
1.  Parm and Float produce identical q,  expected values, logL and rmse.  As expected.
Plots show that the devs are centered on zero.

[...]

PR #544, opened 2023-12-21
revise-code-for-catchability-with-bias-correction

Rick-Methot-NOAA:
* Resolves issue #535

## What tests have been done? 
<-- - [] Test files are in the issue. -->

needs close look by a math-stat whiz

revision for T-dist

### DIFF
--- a/SS_objfunc.tpl
+++ b/SS_objfunc.tpl
@@ -18,6 +18,7 @@
 FUNCTION void evaluate_the_objective_function()
   {
   surv_like.initialize();
+  Svy_like_I.initialize();
   Q_dev_like.initialize();
   disc_like.initialize();
   length_like.initialize();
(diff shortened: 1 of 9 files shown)


SS_objfunc.tpl
```diff
@@ -96,69 +97,56 @@ FUNCTION void evaluate_the_objective_function()
...
           {
-            if (Svy_use(f, i) > 0)
-            {
-              surv_like(f) += 0.5 * square((Svy_obs(f, i) - Svy_est(f, i)) / Svy_se_use(f, i)) + sd_offset * log(Svy_se_use(f, i));
+            if (Svy_errtype(f) == 0) {  // lognormal
+              Svy_like_I(f,i) = 0.5 * square((Svy_obs_log(f, i) - Svy_est(f, i)) / Svy_se_use(f, i))
+                              + sd_offset * log(Svy_se_use(f, i));
+             }
+            else if (Svy_errtype(f) == 1) {  // lognormal with bias adjustment
+              Svy_like_I(f,i) = 0.5 * square((Svy_obs_log(f, i) - Svy_est(f, i) + 0.5 * Svy_se_use(f, i) * Svy_se_use(f, i) ) / Svy_se_use(f, i))
```
Cole-Monnahan-NOAA, line 136: Why not use `square(Svy_se_use(f,i))`?

SS_objfunc.tpl
```diff
@@ -96,69 +97,56 @@ FUNCTION void evaluate_the_objective_function()
...
-          for (i = 1; i <= Svy_N_fleet(f); i++)
+        for (i = 1; i <= Svy_N_fleet(f); i++)
+          if (Svy_use(f, i) > 0)
           {
-            if (Svy_use(f, i) > 0)
-            {
-              surv_like(f) += 0.5 * square((Svy_obs(f, i) - Svy_est(f, i)) / Svy_se_use(f, i)) + sd_offset * log(Svy_se_use(f, i));
+            if (Svy_errtype(f) == 0) {  // lognormal
+              Svy_like_I(f,i) = 0.5 * square((Svy_obs_log(f, i) - Svy_est(f, i)) / Svy_se_use(f, i))
+                              + sd_offset * log(Svy_se_use(f, i));
```
Cole-Monnahan-NOAA, line 133: I don't follow this sd_offset component.

SS_objfunc.tpl
```diff
@@ -96,69 +97,56 @@ FUNCTION void evaluate_the_objective_function()
...
+             }
+            else if (Svy_errtype(f) == 1) {  // lognormal with bias adjustment
+              Svy_like_I(f,i) = 0.5 * square((Svy_obs_log(f, i) - Svy_est(f, i) + 0.5 * Svy_se_use(f, i) * Svy_se_use(f, i) ) / Svy_se_use(f, i))
+                              + sd_offset * log(Svy_se_use(f, i));
+              }
+            else if (Svy_errtype(f) >= 1000) {  // T-dist
```
Cole-Monnahan-NOAA, line 139: Shouldn't this be >1? What if someone used a t-dist with df=4 before. Won't this break it?

SS_objfunc.tpl
```diff
@@ -96,69 +97,56 @@ FUNCTION void evaluate_the_objective_function()
...
+              dvariable df = Svy_errtype(f) - 1000.;
+              Svy_like_I(f,i) = ((df + 1.) / 2.) * log((1. + square((Svy_obs_log(f, i) - Svy_est(f, i))) / (df * square(Svy_se_use(f, i)))))
+                              + sd_offset * log(Svy_se_use(f, i));
+              }
+            else if (Svy_errtype(f) == -1) {  // normal
+              Svy_like_I(f,i) =  0.5 * square((Svy_obs(f, i) - Svy_est(f, i)) / Svy_se_use(f, i))
+                              + sd_offset * log(Svy_se_use(f, i));
             }
           }
```
Cole-Monnahan-NOAA, line 148: Maybe this is caught earlier but what if an invalid Svy_errtype is entered? Say a -5? It looks like it'll just skip the calculations but really should error out or warn?

SS_readdata_330.tpl
```diff
@@ -678,7 +678,7 @@
 
   data_type = 1; //  for surveys
   echoinput << "Units:  0=numbers; 1=biomass; 2=F; >=30 for special patterns" << endl;
-  echoinput << "Errtype:  -1=normal; 0=lognormal; >0=T" << endl;
+  echoinput << "Errtype:  -2=gamma(future); -1=normal; 0=lognormal ; 1=lognormal w/ biascorr; >1XXX=T-dist with DF=XXX" << endl;
```
Cole-Monnahan-NOAA, line 681: df=1 is a cauchy which probably we never want to use, so why not let Errtype>1 be t-dist with df=Errtype?

Cole-Monnahan-NOAA: Actually a potentially easier implementation to leave the t-dist alone is to just use -2 for lognormal bias corr and I'll use -3 for gengamma later. Then you don't have to change anything related to t-dist, I think...

Cole-Monnahan-NOAA (approved):
The equations check out, although I must admit I did get a bit lost in the float options. But I feel confident the lognormal with bias correction and Q as a parameter is correct.

My minor comments are mostly about how this may break backward compatibility with t-distributions in use already. 

Edit: I should add that I was also convinced by the performance in the test case in [this comment](https://github.com/nmfs-ost/ss3-source-code/issues/535#issuecomment-1865380380).

Rick-Methot-NOAA:
Thanks Cole for the good suggestions.  I originally was going to retain T-dist exactly as you suggest.  The reason I went with the distribution you reviewed is that it allowed for options >=0 to use log(surveyobs) and all options <0 would be in normal space.  I have never seen anyone use the T-dist (and wish we had a database of all applications so I would know!) , so I felt that the risk was low.  But the risk is not zero, so I will consider you recommendation.

iantaylor-NOAA:
I found an email from 2020 in which Max Cardinale (@akatan999) referenced using the t-distribution for index likelihoods. It came up because the index plots in r4ss were not working correctly for t-distributed indices. I think the bug was present for at least 6 months and since nobody else reported the issue during that time, it's possible that nobody else was using the t-distribution feature.

Rick-Methot-NOAA:
Thanks Ian.  I will maintain backward compatibility.

SS_objfunc.tpl
Rick-Methot-NOAA: I will change such that

kellijohnson-NOAA:
We are attempting to use the t-distribution for the age-1 index of Pacific Hake abundance in the 2024 stock assessment per your recommendations @Rick-Methot-NOAA when I talked to you last year about trying to fit estimates of zero.

kellijohnson-NOAA:
Also, regarding my review of this code. I am right in the middle of an assessment so if you need my review it will have to wait until at least next week. Sorry.

Cole-Monnahan-NOAA (approved):

Cole-Monnahan-NOAA (approved):

Issue #319, opened 2022-05-23
can the ss3 exe be named something other than ss.exe or ss when running MCMC?


k-doering-NOAA:
This issue is now resolved on Dec 21, 2023 by adding a command line argument invoked by the word "modelname".
For example:
```
ss3.exe -nohess -stopph 3 modelname ss4you
produces output like:
ss4you.par
```
The default value for modelname is "ss3" so this feature is only used if user wants a different name.

Note that if the user already has a par file, it may be named ss.par or ss3.par.  It will need to be renamed if modelname does not match the name of the existing par file.

Original inquiry:
I think this question came up at some point. @kellijohnson-NOAA , perhaps you know that answer to this? Can you use e.g., ss_win.exe or ss_lin  instead as the name for the stock synthesis executable when running MCMC?

kellijohnson-NOAA:
You have a **GREAT** memory @k-doering-NOAA. I did mention this a while ago and I cannot remember what version of SS3 I was using at the time or the name of the file that it said it was missing. But, I just tested it with the current version and everything appears to be working fine even if I name the executable something other than ss.exe.

kellijohnson-NOAA:
Reopening this issue because I finally remembered what was going haywire. When you run Stock Synthesis in mcmc mode you can do so with the executable being named anything you want. But, when you go to do -mceval it thinks that the `.psv` should ALWAYS be named `ss.psv` regardless of what the executable is named. @Rick-Methot-NOAA I think this is an ADMB thing not a SS3 thing, correct? Should I work with @johnoel to see what is going wrong here?

Rick-Methot-NOAA:
Kelli,
the "ss" is coded near top of ss_param.tpl.
I can change that to ss3

kellijohnson-NOAA:
Is there anyway to create it dynamically from the name of the executable rather than have it hard coded? Because as an example, I am currently using an executable called `ss3_2024.exe`, which then creates `ss3_2024.psv` but when I run mceval it is expecting `ss.psv`. So, even if you change the code in ss_param.tpl from ss.psv to ss3.psv I still have to write/do something to change `ss3_2024.psv` to the file name hard coded in ss_param.tpl.

Rick-Methot-NOAA:
I do not think that is possible.  I did not anticipate a workflow as you describe.  I expected users to put ss3.exe in a folder called ss3_2024, or something like that.

kellijohnson-NOAA:
[...]

Issue #56, opened 2020-11-05
SS windows exe since 3.30.14 saves a .log file in directory


k-doering-NOAA:
Imported from redmine, Issue [#70648](https://vlab.ncep.noaa.gov/redmine/issues/70648)
 Opened by @k-doering-NOAA on 2019-10-30
Status when imported: Assigned
 
I've noticed lately (since 3.30.14) that when I run SS in a model folder and SS is either in a different folder or in my path, that SS saves a .log file with the same name as the executable in the folder with the executable (e.g., is called ss.log if the exe is ss.exe). I just checked and this was not the case with 3.30.13 and a few earlier exes. 

Any ideas on what is going on here? I'm guessing @richard.methot and @ian.taylor may know? I was thinking it might be because we changed around the .bat file for compiling recently, but I'm not sure. I'm assuming this is an admb file, but I also am not sure about that.

This is not an urgent issue, but it does clutter up the SS executable folder that I added to my path. It would be nice if this file weren't produced with the executable, but were produced with the other model files or not at all.

PR #545, opened 2023-12-21
allow custom output filenames

Rick-Methot-NOAA:
Resolves issue #319 

## What tests have been done? 
ss3.exe -nohess -stopph 3  modelname ss4you

produces expects to read ss4you.par and will write ss4you.par

test with mcmc

No required change.  Provides option to change filename prefix at command line.
This needs to be documented in command line section of the manual

A new command line argument will set the filename for all ADMB output files to "base_modelname.[ext]"
 where base_modelname can be read from command line with command: modelname followed by the desired name
 e.g.  ss3_win.exe modelname ss4you
When readparfile is used, SS3 will attempt to read ss4you.par.
ss4you will be used as prefix for output files such as ss4you.par and ss4you.cor
if requested modelname.par is not found, then will attempt to read from ss3.par, then ss.par, then will fail with warning.
Whatever name is read, the write will be to modelname.par.
Modelname defaults to ss3.par unless changed by the modelname command.
Which has default of ss3.par.
So, if modelname option is not used and the model reads from ss.par, it will write to ss3.par.
Use of modelname ss will allow writing to ss.par for backward compatibility.

### DIFF
--- a/SS_param.tpl
+++ b/SS_param.tpl
@@ -18,15 +18,35 @@ PARAMETER_SECTION
   // set the filename to all ADMB output files to "base_modelname.[ext]"
   //  where base_modelname can be read from command line with command modelname followed by text
   //  e.g.  ss3_win.exe -nohess -stopph 3  modelname ss4you
+  //  if requested modelname.par is not found, then will attempt to read from ss3.par then ss.par
+  //  whatever name is read, the write will be to modelname.par.  Which has default of ss3.par
   ad_comm::adprogram_name = base_modelname;
   echoinput << "Begin setting up parameters" << endl;
   cout << "Begin setting up parameters ... ";
   if (readparfile >= 1)
   {
-    cout << " read parm file" << endl;
     anystring = base_modelname + ".par";
+    cout << " read parm file" << endl;
+
+    ifstream fin(anystring);
+    if(fin.fail() ) {
+      cout << " no find, try ss3.par" << endl;
+      anystring = "ss3.par";
+      ifstream fin(anystring);
+      if(fin.fail() ) {
+      cout << " no find, try ss.par" << endl;
+      anystring = "ss.par";
+      ifstream fin(anystring);
+      if(fin.fail() ) {
+    	  warnstream << "could not find requested parfile " << anystring;
+    	  write_message(FATAL, 0);
+      }
+    }}
+    cout << " found "<<anystring<<endl;
+
     ad_comm::change_pinfile_name(anystring);
   }
+
   maximum_function_evaluations.allocate(func_eval.indexmin(), func_eval.indexmax());
   maximum_function_evaluations = func_eval;
   convergence_criteria.allocate(func_conv.indexmin(), func_conv.indexmax());
--- a/SS_param.tpl
+++ b/SS_param.tpl
@@ -15,14 +15,17 @@ PARAMETER_SECTION
 //  SS_Label_Info_5.0.1 #Setup convergence critera and max func evaluations
  LOCAL_CALCS
   // clang-format on
-  // set the filename to all ADMB output files to "ss.[ext]"
-  ad_comm::adprogram_name = "ss";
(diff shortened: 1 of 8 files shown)


e-perl-NOAA:
@Rick-Methot-NOAA do you now HAVE to give it modelname each time you run the model or is the default ss3?

Rick-Methot-NOAA:
default is ss3.
Which means it will not read an existing ss.par.

Rick-Methot-NOAA:
Ahh.  I see GHA failing because trying to read the new default ss3.par but the file is named ss.par.  I am inclined to leave this one on the user rather than add a check for existence of ss3.par.  OK?

e-perl-NOAA:
Had to make changes to the ss3-test-models repo to get GHAs passing. Please see PR [#56](https://github.com/nmfs-ost/ss3-test-models/pull/56) in the test-models repo. I will need to make a change in the workflows to go back to using the main branch after that PR is approved and merged. I used the branch I made changes to just to make sure the actions would run here.

e-perl-NOAA:
So does `modelname ss4you` also tell it to read ss4you.par rather than ss3.par in addition to creating ss4you.par?

Rick-Methot-NOAA:
Relying to Elizabeth:  yes.  Whatever the value is for modelname would be used for both reading and writing the .par file.

e-perl-NOAA:
Okay, then with this change should we change the user manual and other documents to say 'ss3.par' instead of 'ss.par'? I also think this will be a huge change for some people. I will create a draft PR of my changes for you to initially look at.

Rick-Methot-NOAA:
good point.  we could keep default at ss.par and anyone could change to ss3.par using the modelname command line argument

e-perl-NOAA:
I guess the difficulty is if the `modelname` specification is also needed to read the old ss.par file instead of the new default ss3.par file.

iantaylor-NOAA:
I think the challenge here is that ss3.par will be simpler and more consistent going forward, it just requires a change for some folks who might have written some scripts or other tools based on the old name. However, if you try to use the par file and it doesn't have the right name, SS3 exits with the message "Error trying to open parameter input file ss.par" (in the future it would say "ss3.par"). I think that error (combined with a comment in the release notes) should sufficient to point folks in the right direction.

Rick-Methot-NOAA:
Let me look into an approach like: if_exist ss.par, then read it, else look for modelname.par

e-perl-NOAA:
@iantaylor-NOAA This error is occurring locally too, it seem to be an error in the SS_plots function and the get_DM_sample_size sub-function within that. The error I am getting locally which is the same one that I see in the github action run is `Error in if (CompError == 1) { : missing value where TRUE/FALSE needed`

SS_param.tpl
```diff
@@ -15,15 +15,38 @@ PARAMETER_SECTION
...
+    if(fin.fail() ) {
+      cout << " no find, try ss3.par" << endl;
+      anystring = "ss3.par";
+      ifstream fin(anystring);
+      if(fin.fail() ) {
+      cout << " no find, try ss.par" << endl;
+      anystring = "ss.par";
+      ifstream fin(anystring);
+      if(fin.fail() ) {
+    	  warnstream << "could not find requested parfile " << anystring;
```
iantaylor-NOAA, line 41: It looks like this error will always report the last attempted filename (ss.par) rather than the one that was requested ([base_modelname].par).

iantaylor-NOAA:
It looks like test-r4ss-with-ss3 is now passing thanks to the changes in https://github.com/r4ss/r4ss/pull/899.

e-perl-NOAA:
Thanks @iantaylor-NOAA!

SS_param.tpl
iantaylor-NOAA: Maybe even better would be to say it could not find either the requested parfile, nor either ss3.par or ss.par, so users know it is checking all three names.

e-perl-NOAA (approved):

PR #546, opened 2023-12-28
Add bootstrap workflow

e-perl-NOAA:

Add workflow to test bootstrap files. Will need to wait on this until the PR to add Simple_with_DM_sizefreq has gone through in the test models repo.

Local testing plus test workflow with push.

### DIFF
--- a/.github/workflows/run-ss3-bootstrap.yml
+++ b/.github/workflows/run-ss3-bootstrap.yml
@@ -77,74 +77,74 @@ jobs:
 
           # Change to have model output 8 bootstrap files
           boot_files <- r4ss::SS_read(dir = boot_mod_dir)
-          # boot_files$start$N_bootstraps <- 10
-          # r4ss::SS_write(boot_files, dir = boot_mod_dir)
-          # r4ss::run(dir = boot_mod_dir, exe = file.path("test-models-repo", "model_runs", "ss3")
+          boot_files$start$N_bootstraps <- 10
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA:
@e-perl-NOAA, I see this is marked as "Draft", but I took a look anyway and it looks good to me. 
It would be good to see it in action, but maybe merging the change is the easiest way to do that. 
To see the output from the `message()` commands, would you just look at the log of the github action? It looks like there's no artifact created, which is fine.

e-perl-NOAA:
Yes, I have it as a draft just because I didn't know if there were any more tests that you could think of that we should add. No artifacts are created. I did run the action from the branch which you can see [here](https://github.com/nmfs-ost/ss3-source-code/actions/runs/7351025171).

iantaylor-NOAA:
Thanks @e-perl-NOAA. The messages in the action run on the branch are easy to follow, but as you say they can be ignored if the action is passing. I can't think of any additional tests, so I would just merge this one and we can always add more in the future if we think of anything.

iantaylor-NOAA (dismissed):

e-perl-NOAA:
@iantaylor-NOAA testing if GitHub support's suggestion of closing the PR and re-opening will allow you to re-review when a commit is made after an approval

iantaylor-NOAA (approved):
Looks good. Testing if this review after close/open works.

e-perl-NOAA:
Okay great! At least the suggestion from the GitHub support team works to get around the bug!

Issue #215, opened 2021-10-13
augment report of overall comps to include breakdown by season and sex


Rick-Methot-NOAA:
New output has these fields after adding area and season:
OVERALL_COMPS report:30
area seas Fleet N_obs len_bins  10 14 18 22 etc.

New rows include:
season = 0 to report combined seasons but only if nseas>1
rows for males and females in a two sex model.

Augmented report working for length comps; need to make comparable changes to age comp summary.

PR #555, opened 2024-01-19
add warning when addtocomp is not > 0

iantaylor-NOAA:
A "nan" likelihood in model shared by @akatan999 took some time to debug (traced to age comp likelihoood, then to rows with zero observations for some bins, then to 0 values for `addtocomp` settings).
It's hard to imagine anyone wanting the `addtocomp` to be zero, so adding a warning when this occurs will make debugging this easier.

## What tests have been done? 
Compiled and run on the model that was shared which now produces these warnings:
```
Warning 1 : addtocomp input for age comps fleet 1 is 0 but should be > 0 to avoid log(0) in likelihood
Warning 2 : addtocomp input for age comps fleet 2 is 0 but should be > 0 to avoid log(0) in likelihood
```
Can be replicated on any model by changing `addtocomp` for length or age data to 0 or negative value.

Review by @Rick-Methot-NOAA to confirm that this was done correction and should be a warning and not an stop. Also, it could be possible to only warn if there are zero values in the observed or expected comps after addition of `addtocomp` but that would be more challenging to implement. Lastly, I'm not clear on the second argument to `write_message(WARN, 0)`, so would be good to get confirmation that this shouldn't be `write_message(WARN, 1)` instead.

### DIFF
--- a/SS_readdata_330.tpl
+++ b/SS_readdata_330.tpl
@@ -1402,6 +1402,11 @@
         min_tail_L(1, f) = min_tail_L(0, f);
         min_tail_L(2, f) = min_tail_L(0, f);
         *(ad_comm::global_datafile) >> min_comp_L(0, f);
+        if (min_comp_L(0, f) <= 0) {
+          warnstream << "addtocomp input for length comps fleet " << f << 
+          " is " << min_comp_L(0, f) << " but should be > 0 to avoid log(0) in likelihood";
+          write_message(WARN, 0);      
+        }
         min_comp_L(1, f) = min_comp_L(0, f);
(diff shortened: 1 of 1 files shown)


Rick-Methot-NOAA (approved):
Looks good.  Can we provide some guidance in documentation on implications of how much bigger than >0 they should use?

e-perl-NOAA:
See PR [#215](https://github.com/nmfs-ost/ss3-doc/pull/215) in ss3-doc repo.

e-perl-NOAA:
@iantaylor-NOAA do you mind if I go ahead and merge this?

PR #552, opened 2024-01-10
Retry ADMB docker images

johnoel:
Retry docker targets.

PR #568, opened 2024-02-13
Update workflows admb docker

e-perl-NOAA:

Update all workflows to use the admb docker image.

See PR #552 


e-perl-NOAA:
@Rick-Methot-NOAA please look at the new warning_ss_ref.txt file. There are some warnings produced by the admb docker image that we haven't seen prior but look like they might be easy fixes.

Rick-Methot-NOAA:
I saw one warning regarding an unguarded "For".  I will fix that.

Note that For statements that rely on indentation for interpretation are easily broken.  I try to always use explicit {}.

e-perl-NOAA:
@iantaylor-NOAA my thought process for retaining them is to have them easily accessible in case 1) the docker image stops working and 2) we need to test compiling locally across all/multiple OS' (in the case that admb no longer will compile SS3 locally and we need to see if it's a compiler issue or if we need to see if it will compile locally when we finally get updated to windows 11).

Rick-Methot-NOAA:
Surely someone is already compiling with Win11.  Is there significant risk?

e-perl-NOAA:
I'm not sure at this point. I can delete the commented-out portion that downloads admb and puts it in the path for most of the workflows but leave it in one (probably the build-ss3) just to have easily accessible as a *just in case* measure, at least until we have a better understanding of what complications could arise?

Rick-Methot-NOAA:
I think it makes sense to archive it in some readily retrievable form.  The source code repo has lots of commits, so if we rely on git history to retrieve this, then we at least should have a label on that commit.

e-perl-NOAA:
Okay sounds good. I'll go ahead and merge then!

Issue #570, opened 2024-02-20
[Bug]: FMethod 4 not defaulting to hybrid


Rick-Methot-NOAA:
### Describe the bug

FMethod 4 should not require that hybrid fleets be specified, they should be default.  But an user example was assigning them F=0, not F=hybrid.  Need to investigate

### To Reproduce

With FMethod 4, leave off an entry for a fishing fleet.

### Expected behavior

dedault to hybrid for unspecified fishing fleets

### Which OS are you seeing the problem on?

Windows

### Which version of SS3 are you seeing the problem on?

3.30.22.01

### Additional Context

Nicholas' shark model

PR #577, opened 2024-02-29
570 simple fix for FM4

Rick-Methot-NOAA:
resolves issue #570 
more complete work on Fmethod will follow in the unify F approach branch


iantaylor-NOAA (approved):
Looks good.
Looking through the SS_readcontrol_330.tpl, I think I now understand the logic of how F_Method_PH is set and used and the change makes sense.

Issue #569, opened 2024-02-19
[Bug]:  discard-at-age output missing for forecast years when using age-specific discard


Rick-Methot-NOAA:
### Describe the bug

 discard-at-age output missing for forecast years when using age-specific discard

### To Reproduce

see output from Max's north sea sole

### Expected behavior

the output

### Which version of SS3 are you seeing the problem on?

3.30.22.01

### Additional Context

[age_discard.zip](https://github.com/nmfs-ost/ss3-source-code/files/14335422/age_discard.zip)

PR #580, opened 2024-03-13
Unify F approach

Rick-Methot-NOAA:
continues development for issue #570  and especially #581
See description and example in #581
Substantial refactoring of the internal machinery for managing F as parameter vs F as hybrid
extends the F-detail approach previously available only for F_method 2 to F_method 4

only new input is for F_method 4 if using the F_detail feature.

### DIFF
--- a/SS_global.tpl
+++ b/SS_global.tpl
@@ -902,18 +902,20 @@ BETWEEN_PHASES_SECTION
     last_objfun = obj_fun;
   }
 
-  //  SS_Label_Info_11.2 #For Fmethod=2 & 4, set parameter values (F_rate) equal to Hrate array fromcalculated using hybrid method in previous phase
+  //  SS_Label_Info_11.2 #For Fmethod=2 & 4, set parameter values (F_rate) equal to Hrate calculated using hybrid method in previous phase
   if (N_Fparm > 0 && j_phase > 1)
   {
     for (int ff = 1; ff <= N_catchfleets(0); ff++)
     {
       f = fish_fleet_area(0, ff);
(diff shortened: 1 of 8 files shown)


e-perl-NOAA:
Okay so the spinydogfish model echoinput file is stopping after:
```
GP Seas Area Use? female_recr_dist
1 1 1 1 0.5
 made the age_age' key 
Finished evaluating biology calculations once


Begin estimating
```

Warning file looks like the following:
```
Note 1 Information: Max data length bin: 132  < max pop len bins: 136; so will accumulate larger pop len bins
3 iter 1
 hrate for 1970: 
```

Rick-Methot-NOAA:
that stop in dogfish might only be because I had not yet removed a specific output statement while debugging.  It is gone now.

e-perl-NOAA:
Yes, all seems to be working!

iantaylor-NOAA (approved):
Looks good. I didn't try to follow all the details, but I don't see any red flags. The tests are all passing, including the vermillion_snapper_F4 model which uses F Method 4. 

I compared Report files and see the DISCARD_AT_AGE is now available in the forecast as intended (discussed in #569). 
image

I see control.ss_new remains the same except for improved comments
image

Rick-Methot-NOAA:
Thanks Ian.  Nothing magic about the 2.9 value.  The role of max F is mostly to keep model from exploring F levels that nearly extirpate older fish.  In the attached example, it is constraining during early hybrid phases to keep F below 2.9, which creates a catch penalty, which moves model back to higher biomass and lower F values.  For long-lived stocks a maxF of 0.4 might seem reasonable, but even 0.40 will not extirpate.

PR #582, opened 2024-03-19
refine echoinput messages and errors in Q_setup

iantaylor-NOAA:
Refines echoinput messages and errors associated with read of Q_setup in control file inspired by recent forums post by Laura Lee: https://vlab.noaa.gov/web/stock-synthesis/public-forums/-/message_boards/message/35975358.
* makes more clear which fleet has an issue in Q setup
* aligns notation (now just "Q setup" instead of also "q setup" and "Qsetup")
* adds explanation for why the echoed Q setup table looks different than input
* adds explanation for Q_Npar and Q_Npar2
 
Files attached to forums messages were used for testing.

Make sure the refinements make sense.

### DIFF
--- a/SS_readcontrol_330.tpl
+++ b/SS_readcontrol_330.tpl
@@ -2944,7 +2944,7 @@
   // 5:  0/1 to float
   
   //  read setup and get the parameter count
-  echoinput << "# read Q setup only for fleets with survey/CPUE/effort observations, end with fleet_ID<0 "
+  echoinput << "# read Q setup only for fleets with survey/CPUE/effort observations, end with fleet_ID<0 " << endl
             << "#  fleet_ID link_type link_info  extra_se(0/1)   biasadj(0/1)  float(0/1)" << endl;
   firstQparm = ParCount; //  base index before adding Q parms
   Q_setup.initialize();
@@ -2965,7 +2965,9 @@
       *(ad_comm::global_datafile) >> tempvec(1, 5);
     }
   } while (j > 0);
-  echoinput << "q setup " << endl
+  echoinput << "Q setup " << endl
+            << "Note: control file should only contain rows for fleets with index observations" << endl
+            << "      but setup is reported here as one row for each fleet (with no fleet column)" << endl
             << Q_setup << endl;
   
   //  get base parameter count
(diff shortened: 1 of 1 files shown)


SS_readcontrol_330.tpl
```diff
@@ -2944,7 +2944,7 @@
   // 5:  0/1 to float
   
   //  read setup and get the parameter count
-  echoinput << "# read Q setup only for fleets with survey/CPUE/effort observations, end with fleet_ID<0 "
+  echoinput << "# read Q setup only for fleets with survey/CPUE/effort observations, end with fleet_ID<0 " << endl
```
e-perl-NOAA, line 2947: Should all lines have a ";" at the end? I see that most lines do but not all of them so I'm not sure if it has to have it or should have it/

SS_readcontrol_330.tpl
```diff
@@ -3071,15 +3073,15 @@
...
     }
   }
-  
-  echoinput << "q setup " << endl
-            << Q_setup << endl;
-  echoinput << "q setup parms " << endl
+
+  echoinput << "Q_Npar (number of long parameter lines): " << Q_Npar << endl
+            << endl << "Q setup parameter index " << endl
+            << "column have index of first parameter for:  1=base q with link;  2=extrastd; 3=env; 4=block/trend; 5=dev;"
```
e-perl-NOAA, line 3084: Should there be " << endl" here? Also, same question about ";" at the end of lines here.

Rick-Methot-NOAA:
I am not aware of any past discussion on the need for consistency with regard to punctuation at the end of such lines in the output

e-perl-NOAA:
Okay, great! Those were just questions I had based on not having a firm knowledge of C++ and if things of that sort would cause issues down the road.

Rick-Methot-NOAA:
Ahh.  I see that you were referring to the C++ structure.  A line not ending with ; is just continued to the next line.

Rick-Methot-NOAA (approved):

iantaylor-NOAA:
@e-perl-NOAA I didn't even think about the semi colons,  so good question.

I was unconsciously following the style added by the automated code-styling Neil applied in 2022 (discussed in https://github.com/nmfs-ost/ss3-source-code/issues/163), which added lines breaks after every `endl`. For instance, commit 6252695 converting things like
```
SS2out<<endl<<"Starter_Comments"<<endl<<Starter_Comments<<endl;
```
to
```
    SS2out << endl
           << "Starter_Comments" << endl
           << Starter_Comments << endl;
```

There's still just the one semi-colon at the end, but line breaks within it to improve readability.
That code is equivalent to this version with 3 semi-colons:
```
    SS2out << endl;
    SS2out << "Starter_Comments" << endl;
    SS2out << Starter_Comments << endl;
```

Issue #579, opened 2024-03-12
Extending time-blocks into forecast years


IJStewart:
### Describe the bug

The 3.30.22.1 manual recommends use of "9999" as the end year of the terminal time block in order to ensure it will extend to cover the entire forecast period (page 217). However, when this approach is used, a warning is generated "...9999 reset to YrMax: XXXX". As this is a recommended approach, I don't think it should generate a warning.

### To Reproduce

Set the end of the last time block to "9999".

### Expected behavior

Extension of the block to the end of the forecast period without a warning.

### Which OS are you seeing the problem on?

Windows

### Which version of SS3 are you seeing the problem on?

3.30.22.1

Rick-Methot-NOAA:
Thanks Ian.  I will revise the code such that if the user enters 9999, SS3 will not generate a warning.

PR #583, opened 2024-03-20
#579 update block warning

Rick-Methot-NOAA:
This revision to the warning on block endings will resolve issue #579 

none


iantaylor-NOAA (approved):
Looks good.

PR #585, opened 2024-03-28
Build admb and ss3 from source github action

e-perl-NOAA:

- Adding a github action that that is only triggered manually to build admb and ss3 from source for windows, mac, and linux. This action is a backup for if/when the admb docker images don't work. Building both admb and ss3 from source gives us more control over the g++ libraries too. [Successful GitHub actions run of this workflow](https://github.com/nmfs-ost/ss3-source-code/actions/runs/8457542489).
- For windows, admb is specifically built using the compilers in Rtools 42.
- Also updated the actions/checkout to version 4 for all workflows

None.

### DIFF
--- a/.github/workflows/build-admb-and-ss3-from-source.yml
+++ b/.github/workflows/build-admb-and-ss3-from-source.yml
@@ -34,7 +34,7 @@ jobs:
           rtools-version: '42'
           windows-path-include-rtools: TRUE
           update-rtools: TRUE
-          
+
     # Checkout ADMB repository
       - name: checkout admb and build from source, windows
         uses: actions/checkout@v4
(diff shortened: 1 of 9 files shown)


.github/workflows/build-admb-and-ss3-from-source.yml
```diff
@@ -0,0 +1,290 @@
+# Builds the stock synthesis executables for testing and distribution.
+# for mac, linux, windows
+# Runs on every push and PR (even draft PRs)
+name: build-admb-and-ss3-from-source
+
+on:
+    workflow_dispatch:
+
```
iantaylor-NOAA, line 8: Could we add a schedule here so that it runs weekly or monthly so we figure out if anything breaks before we need to rely on it?
```  
schedule:
    - cron: '0 15 * * 2' // every Tuesday at 3pm UTC (= 10am or 11am Eastern time)
```
I think that is right based on syntax found here:https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule)

iantaylor-NOAA (approved):
Looks good. 
And good idea to have this backup. 

I notice that it doesn't take that much longer than the regular build-ss3 workflow (28 minutes vs 17-20 minutes), but I think it still makes sense to stick with the docker workflow with this one as a backup just in case we need it.

Rick-Methot-NOAA (approved):
Looks good to me.

e-perl-NOAA:
@iantaylor-NOAA it doesn't take significantly longer because of the make -j 4 command which allows admb to be built in parallel.

PR #587, opened 2024-04-16
Build macos 14

e-perl-NOAA:

The M2 chip macs that use arm64 (including macos-14) seem to behave a bit differently with the ss3 executables than the macs that use x86_64 architecture (macos-12 and before). Building the executable on a macos-14 runner fixed the issue that a user (Juan-Carlos) was having. I have now included building on a macos-14 machine in our GitHub actions. @iantaylor-NOAA noted that the macos-14 built executable does not work on older macs so building macos-12 is also necessary.

**In theory**, the [macos-latest image is supposed to be macos-14 right now](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) BUT a [recent run](https://github.com/nmfs-ost/ss3-test-models/actions/runs/8726783320/job/23942823321#step:8:14) testing updates I'm making to the r4ss get_ss3_exe() function to get the m2 exe showed that it had an x86_64 architecture OS (see R.version output) which is NOT what macos-14 had, so I'm not sure what's up there. When it does sort itself out, i'll switch where it says macos-14 to macos-latest. 

The macos-12 runner will eventually be discontinued. The macos-13 runner does not create an exe that will work when built with the workflow that uses the docker image (despite 2 days of trying) but it will create an exe that will work when built using the build-admb-and-ss3 workflow. 

I added a mac version that works for the arm64 architecture onto the v.3.30.21.1 release. 

Also officially updated to admb 13.2 version for github actions.

None.

### DIFF
--- a/.github/workflows/build-admb-and-ss3-from-source.yml
+++ b/.github/workflows/build-admb-and-ss3-from-source.yml
@@ -25,6 +25,8 @@ jobs:
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v4
+        with:
+          ref: v3.30.22.1
     
     # Set up R
       - name: Set up R, specify rtools version and path for windows
--- a/.github/workflows/build-admb-and-ss3-from-source.yml
+++ b/.github/workflows/build-admb-and-ss3-from-source.yml
@@ -16,7 +16,7 @@ jobs:
       matrix:
         config:
           - {os: windows-latest}
-          - {os: macos-14}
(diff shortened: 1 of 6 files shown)


e-perl-NOAA:
See [successful github actions runs on the try-build-macos-14 branch](https://github.com/nmfs-ost/ss3-source-code/actions).

iantaylor-NOAA (approved):
Thank you @e-perl-NOAA!

Issue #586, opened 2024-04-11
[Feature]: 10+ multiyear average F_std_scaling; with change to log(ratio) approach


HaikunXu:
### Describe the solution you would like.
After considering thealternatives listed below, a third approach has been implemented:
120.1 # F_std_scaling: 0=no scaling; 1=F/Fspr; 2=F/Fmsy; 3=F/Fbtgt; where F means annual F_std, Fmsy means F_std@msy; values >=11 invoke N multiyr using 10s and 100s digit; append .1 to invoke log(ratio)
This approach breaks backward compatibility for anyone previously using values >100 to invoke log(ratio), but other proposed approaches also broke compatibility or introduced more complexity to the algorithm.

### Describe alternatives you have considered

2 # F_std_scaling: 0=no scaling; 1=F/Fspr; 2=F/Fmsy; 3=F/Fbtgt; where F means annual F_std, Fmsy means F_std@msy; values >=11 invoke N multiyr (up to 9!) using 10's digit; >100 invokes log(ratio); _**< -100 invoke N multiyr (more than 9) using 100’s and 10's digits to choose the years to average; < -1000 invoke log(ratio)**_

2 # F_std_scaling: 0=no scaling; 1=F/Fspr; 2=F/Fmsy; 3=F/Fbtgt; where F means annual F_std, Fmsy means F_std@msy; values >=11 invoke N multiyr (up to 9!) using 10's digit; >100 invokes log(ratio); **_< -1000 invoke N multiyr (more than 9) using 1000’s and 100's digits (could use more than two digits for 100+ years) to choose the years to average; the 10’s digit informs whether to use log (1) or not (0)_**

### Describe if this is needed for a management application

We need to estimate the std of F/FMSY for a 12-year window for management purposes.

HaikunXu:
Hi Rick,

Your proposed specification looks reasonable to me.

Rick-Methot-NOAA:
- [ ] need to extend same logic to other types of multi-yr averaging
- [ ] multiyr averaging only applies to time_series; should we add a feature such that the averaging is continuous from time_series into forecast?

PR #589, opened 2024-04-19
change multiyr average approach

Rick-Methot-NOAA:
Allows for multiyear averaging of F/Fmsy for >10 years.   Per request from @HaikunXu

* Resolves issue #586

## What tests have been done? 
example model "simple" was used as test case to demonstrate fidelity of the change.

None

in starter.ss there are needed changes if multiyr or log(ratio) is invoked

### DIFF
--- a/SS_readstarter.tpl
+++ b/SS_readstarter.tpl
@@ -640,20 +640,18 @@
   int depletion_basis;
   int depletion_multi;
   int depletion_log;
-  init_int depletion_basis_rd; // 0=skip; 1=B0; 2=Bmsy; 3=B_styr; 4=B_endyr; 5=dynamic_Bzero; values >=11 invoke multiyr with 10's digit; >=100 invoke log(ratio) with hundreds digit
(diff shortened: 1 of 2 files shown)


SS_readstarter.tpl
```diff
@@ -640,20 +640,18 @@
   int depletion_basis;
   int depletion_multi;
   int depletion_log;
-  init_int depletion_basis_rd; // 0=skip; 1=B0; 2=Bmsy; 3=B_styr; 4=B_endyr; 5=dynamic_Bzero; values >=11 invoke multiyr with 10's digit; >=100 invoke log(ratio) with hundreds digit
+  init_number depletion_basis_rd; // 0=skip; 1=B0; 2=Bmsy; 3=B_styr; 4=B_endyr; 5=dynamic_Bzero; values >=11 invoke multiyr with 10's digit; append .1 to invoke log(ratio) with hundreds digit
```
chantelwetzel-noaa, line 643: Can we add an example here of what that would look like for a specific option?

chantelwetzel-noaa: Alternatively, we could include that in the manual if it that would be preferred.

Rick-Methot-NOAA:
@chantelwetzel-noaa  I like your suggestion of putting an example directly in starter.ss_new.

chantelwetzel-noaa (approved):

Issue #595, opened 2024-04-30
[Feature]: Add 2-parm catchability (Q) option with offset and slope


Rick-Methot-NOAA:
### Describe the solution you would like.

When an environmental index is used to compare to a dev vector, including recruitment devs, there is a need for an offset parameter as well as the existing slope parameter.
This will be a step in implementation of #5 

Input looks like:
```
#_Q_setup for fleets with cpue or survey data
#_1:  fleet number
#_2:  link type: 1=simple q; 2=mirror; 3=power (+1 parm); 4=mirror with rescale (+1p); 5=offset (+1p); 6=offset & power (+2p)
#_3:  extra input for link, i.e. mirror fleet# or dev index number
#_4:  0/1 to select extra sd parameter
#_5:  0/1 for biasadj or not
#_6:  0/1 to float
#_   fleet      link link_info  extra_se   biasadj     float  #  fleetname
         2         1         0         0         0         0  #  Survey
         3         5         0         0         0         0  #  env
-9999 0 0 0 0 0
#
#_Q_parameters
#_          LO            HI          INIT         PRIOR         PR_SD       PR_type      PHASE    env-var    use_dev   dev_mnyr   dev_mxyr     dev_PH      Block    Blk_Fxn  #  parm_name
           -20            20             0             0            99             0          3          0          0          0          0          0          0          0  #  LnQ_base_Survey(2)
         0.001            20             1             0            99             0          2          0          0          0          0          0          0          0  #  Q_base_env(3)
            -1             1             0             0            99             0          2          0          0          0          0          0          0          0  #  Q_offset_env(3)

```
Note that this new parameter is now labelled Q_offset.  Previously the Q parameter associated with option 3 (mirror +) was called Q_offset.  That is renamed to Q_scale because it is principally used to scale a CPUE to the  relative area it represents.

### Describe alternatives you have considered

force users to remember to zero-center their env data, but even zero-centering is an incomplete alternative to allowing for an offset parameter.

### Statistical validity, if applicable

bias will be created if a non-zero centered env index is used

### Describe if this is needed for a management application

bias will be created if a non-zero centered env index is used

iantaylor-NOAA:
Good to see this moving forward.

[...]

PR #596, opened 2024-04-30
add offset option for ENV Q

Rick-Methot-NOAA:
* Resolves issue #595 
* is responsive to issue #5

### DIFF
--- a/SS_readcontrol_330.tpl
+++ b/SS_readcontrol_330.tpl
@@ -2985,7 +2985,11 @@
       if (Svy_errtype(f) == -1)
       {
         ParmLabel += "Q_base_" + fleetname(f) + "(" + NumLbl(f) + ")";
-      }
+        if (Q_setup(f, 5) == 1) //  float
+        {
+          warnstream << "Q setup:  can't use float option with normal error distribution, fleet: " << f;
+          write_message (FATAL, 0); // EXIT!
+        }
       else if (Svy_errtype(f) >= 0)  // lognormal or T-dist
       {
         ParmLabel += "LnQ_base_" + fleetname(f) + "(" + NumLbl(f) + ")";
--- a/SS_readcontrol_330.tpl
+++ b/SS_readcontrol_330.tpl
@@ -3044,6 +3044,13 @@
           ParmLabel += "Q_mirror_offset_" + fleetname(f) + "(" + NumLbl(f) + ")";
           break;
         }
+        case 5: //  add offset
+        {
+          Q_Npar++;
+          ParCount++;
+          ParmLabel += "Q_offset_" + fleetname(f) + "(" + NumLbl(f) + ")";
+          break;
+        }
       }
       if (Q_setup(f, 3) > 0)
       {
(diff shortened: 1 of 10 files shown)


SS_readcontrol_330.tpl
```diff
@@ -3052,6 +3059,14 @@
         Q_setup_parms(f, 2) = Q_Npar;
         ParmLabel += "Q_extraSD_" + fleetname(f) + "(" + NumLbl(f) + ")";
       }
+      if (Svy_units(f) == 35 || Svy_units(f) == 36)  // env index of recdev or parm dev vector
+      {
+        if (Q_setup(f, 1) != 5)
+        {
+          warnstream << "Suggest using Q option 5 to include offset parameter for an ENV index";
```
iantaylor-NOAA, line 3066: I would change `for an ENV index` to `for an index of deviations (type 35 or 36)` to be both more general and more explicit on which types the recommendation is for.

iantaylor-NOAA (approved):
Looks good. Thanks for adding this valuable feature. It does have an impact on the petrale model as expected.

The only confusion I have is in how the calculation of expected index value works.
Based on the code, I would have thought that Exp = Q_base * (Vuln_bio + Q_offset), but when I attempt that calculation from the INDEX_2 output in the attached spreadsheet using the example model attached to #596, I get a difference of 0.06549 for all years of the index.

Spreadsheet: [env_index_calcs.xlsx](https://github.com/nmfs-ost/ss3-source-code/files/15172459/env_index_calcs.xlsx)

SS_readcontrol_330.tpl
Rick-Methot-NOAA: The code was done same as for Power function:
```
                if (Q_setup(f, 1) == 5) //  add offset
                {
                  vbio += Q_parm(Q_setup_parms(f, 1) + 1);
                }

                // SS_Label_Info_46.1.1 #note order of operations,  vbio raised to a power, then constant is added, then later multiplied by Q.  Needs work
                if (Q_setup(f, 1) == 3) //  link is power function
                {
                  vbio = pow(vbio, 1.0 + Q_parm(Q_setup_parms(f, 1) + 1));
                } //  raise vbio to a power

                Svy_selec_abund(f, j) = vbio;  //  e.g. the abundance that has been selected through selectivity or other assignment process
```
Which is then presented as available biomass in table Index_2,  Which is confusing.  Better to save selex_abund untransformed (e.g. selectivity only), then expected values will have q applied.  I'll check to see if that works.

Rick-Methot-NOAA: @ian-taylor I think I need to deprecate the output column eff_Q which was simply svy_est / svy_selec_abund.  It sort of worked when svy_selec_abund was vbio because it included effect of the offset.  Now with the offset it is a misleading output.  I think best to deprecate it and put NA in the column.

iantaylor-NOAA: The numbers now line up more intuitively, so thanks for making the change.

The only place that I've used eff_Q was looking at the change in Q as a function of vulnerable biomass (or what I thought was vbio) to see the effect of a power function on catchability. If there's another way to get that information, I'm OK with  eff_Q going away. Just to clarify, would it only go away for links 5 and 6 or for all links (in which case I think the column could be removed).

Rick-Methot-NOAA: report now has NA if offset is used.  Also, values of offset and power now added as new columns for INDEX_1 table, with NA as appropriate.

iantaylor-NOAA: Expanded table looks good. Existing r4ss code showing vulnerable biomass vs. effective catchability continues to work for a model that used power function on Q with lognormal index error. Happy to have this squashed and merged assuming tests pass.

e-perl-NOAA:
@Rick-Methot-NOAA Is there anything that needs to be updated in the documentation for this?

Rick-Methot-NOAA:
@e-perl-NOAA  The google doc labelled catchbility is intended to replace the existing manual text.

PR #608, opened 2024-07-19
make column headings more consistent in data files 

iantaylor-NOAA:
Initially posting as a draft pull request to foster discussion before marking ready for review.

- Aligns a few inconsistencies in the column headers written to data_echo.ss_new (and expected and boostrap data files).
  - more consistent use of lowercase headers
  - "season" to "seas" (which is more common)
  - "partition" to "part"  (which is more common)
  - "SampleSize" to "Nsamp" (which is more common)
  - make things match better across the three types of data ss_new files as well as echoinput (in part because this helped confirm that the changes above were made throughout)
- Change "yr" to "year" as suggested in this comment and the one below from @kellijohnson-NOAA https://github.com/r4ss/r4ss/issues/512#issuecomment-1438728972.
- Changes are in parallel with changes in the r4ss function to read the data files initiated thanks to @e-perl-NOAA in https://github.com/r4ss/r4ss/pull/945.

## What tests have been done? 
None so far. The changes should only be to comments, not numerical values, so I don't think additional testing of SS3 itself is needed.

## What tests/review still need to be done?
The read/write of data files by r4ss should be tested against these changes along with the status-quo output.

### DIFF
--- a/SS_prelim.tpl
+++ b/SS_prelim.tpl
@@ -1424,7 +1424,7 @@
     bodywtout << "# fleet 0 contains begin season pop WT" << endl;
     bodywtout << "# fleet -1 contains mid season pop WT" << endl;
     bodywtout << "# fleet -2 contains maturity*fecundity" << endl;
-    bodywtout << "#Yr Seas Sex Bio_Pattern BirthSeas Fleet " << age_vector << endl;
+    bodywtout << "#_year seas sex bio_pattern birthseas fleet " << age_vector << endl;
   }
 
   if (Turn_off_phase < 0)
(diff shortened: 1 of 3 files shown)


iantaylor-NOAA:
@e-perl-NOAA and I discussed the changes to SS3 and r4ss just now and made a few more tweaks, but we think this is now ready for review from @Rick-Methot-NOAA.
We will also merge the parallel issue cleaning up r4ss column headers: https://github.com/r4ss/r4ss/pull/945.

Rick-Methot-NOAA (approved):
nice cleanup.  thank you

PR #609, opened 2024-07-30
separate annual timeseries from spr series; improve mean_age reporting

### LINKED ISSUE #607: revise SPR_SERIES in Report.sso to clarify
The use of multiple columns in the SPR_SERIES with the same column name ("Bio_all" and "Bio_Smry") with "Actual:" in between can be confusing:
https://github.com/nmfs-ost/ss3-source-code/blob/95ed8550ae26208ca4ff2b146a945d028337b2a4/SS_write_report.tpl#L1575-L1576

I think it may be worth appending something to the columns that don't match what's in TIME_SERIES to clarify the difference.

Rick-Methot-NOAA:
resolves #607 

Information is in the issue.
The annual_time_series table has been separated from the SPR_series table such that only per-recruit values are in the SPR table.  
update and check the mean age values
add more notes on each table

Testing can be done with any model.

needs work with r4ss

### DIFF
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -1414,9 +1414,9 @@ FUNCTION void write_bigoutput()
     {
       for (y = styr - 2; y <= YrMax; y++)
       {
-        if (y <= endyr && p == 1)
+        if (p == 1)
         {
-          Smry_Table(y)(15, 17).initialize();
+          Smry_Table(y)(16, 17).initialize();
         }
         for (s = 1; s <= nseas; s++)
         {
@@ -1447,18 +1447,19 @@ FUNCTION void write_bigoutput()
               smrynum += sum(natage(t, p, g)(Smry_Age, nages));
               smryage += natage(t, p, g)(Smry_Age, nages) * r_ages(Smry_Age, nages);
               SSB_vir_LH += natage(t, p, g) * virg_fec(g);
-              if (y <= endyr)
               {
                 for (f = 1; f <= Nfleet; f++)
                 {
                   if (fleet_area(f) == p && y >= styr - 1 && fleet_type(f) <= 2)
                   {
-                    Smry_Table(y, 16) += sum(catage(t, f, g));
+                    Smry_Table(y, 16) += sum(catage(t, f, g));  // temporary storage spot
                     Smry_Table(y, 17) += catage(t, f, g) * r_ages;
                   }
                 }
               }
             } //close gmorph loop
+          Smry_Table(y, 17) /= ( Smry_Table(y, 16) + 1.0e-06 ); // mean age of catch
+          Smry_Table(y, 16) = smryage / smrynum; //  mean age of summary numbers
           if (gender_rd == -1)
             SSB_vir_LH *= femfrac(1);
           SS2out << p << " " << y;
@@ -1514,10 +1515,6 @@ FUNCTION void write_bigoutput()
           }
           SS2out << " " << Bio_Comp << " " << Num_Comp;
           SS2out << " " << SSB_B_yr(y) << " " << SSB_N_yr(y);
-          if (s == 1 && y <= endyr)
-          {
-            Smry_Table(y, 15) += smryage;
-          } // already calculated for the forecast years
           for (f = 1; f <= Nfleet; f++)
             if (fleet_type(f) <= 2)
             {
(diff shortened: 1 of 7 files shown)


SS_write_report.tpl
```diff
@@ -1560,31 +1557,21 @@ FUNCTION void write_bigoutput()
...
-    SS2out << " Enc_Catch_B Dead_Catch_B Retain_Catch_B  Enc_Catch_N Dead_Catch_N Retain_Catch_N sum_Apical_F F=Z-M  M";
-    SS2out << endl;
+           << pick_report_name(17) << endl;
+    SS2out << "#_NOTE:_reports_per_recruit_quantities_using_current_year_biology;_using_same_equil_calc_routine_used_for_reference_points" << endl;
+    SS2out << "#_NOTE:_current_year_biology_is_current_at-age_biology_from_time_series;_not_recalc_biology_from_current_growth_parameters" << endl;
+    SS2out << "#_NOTE:_uses_R0= " << Recr_virgin << endl;
+    SS2out << "#_NOTE:_Y/R_unit_is_Dead_Biomass" << endl;
+    SS2out << "#_NOTE:_gentime_is_mean_age_of_female_spawners_weighted_by_reproductive_value-at-age_(fec(g))" << endl;
+
+    SS2out << "Yr Era Bio_all_eq Bio_Smry_eq SSB_unfished_eq SSBfished_eq SSBfished/R SPR  Y/R GenTime" << endl;
```
e-perl-NOAA, line 1574: Is there a reason to change YPR to Y/R? It seems like YPR is used most other places.

Rick-Methot-NOAA: True, but I was striving for consistency with SSB/R.  Looked at other instances and agrtee with you, so will change back.

SS_write_report.tpl
```diff
@@ -1597,6 +1584,34 @@ FUNCTION void write_bigoutput()
...
+    // end SPR time series
+
+    SS2out << "#" << endl << "ANNUAL_TIME_SERIES report:17a" << endl;
+    SS2out << "#_NOTE: MnAgeSmry_is_numbers_weighted_meanage_at_and_above_smryage:_" << Smry_Age << endl;
+    SS2out << "#_NOTE:_mean_age_of_catch_is_numbers-weighted_and_based_on_catage_which_is_the_dead_catch_and_comes_from:_sel_dead_num_=_sel_*_(retain_+_(1-retain)*discmort)" << endl;
+    SS2out << "#_NOTE: Depletion_basis: " << depletion_basis << " # " << depletion_basis_label << endl;
+    SS2out << "#_NOTE: F_report_basis: " << F_reporting << " # " << F_report_label << endl;
+    SS2out << "#_NOTE: SPR_report_basis: " << SPR_reporting << " # " << SPR_report_label << endl;
+    SS2out << "year Era Bio_all_an Bio_Smry_an Num_Smry_an SSB recruits sel_catch_B_an dead_catch_B_an retain_catch_B_an sel_catch_N_an dead_catch_N_an retain_catch_N_an" <<
+              " dead_catch_B/bio_smry sum_fleet_F F=Z-M M mn_age_SSB mn_age_smry mn_age_catch  SPR_std Depletion_std F_std" << endl;
```
e-perl-NOAA, line 1598: Is sum_fleet_F the same as Tot_Exploit?

Rick-Methot-NOAA: different.  But that raises topic, which I m considering:
a.  should I add a note to describe every cryptic column header?
b.  what is preferred format of such notes:
1.  before or after the table
2. special code to denote that the next line will be start of a table
3. text uses spaces, or spaces all filled with a character like "_"

e-perl-NOAA: I think it would help. I think before the table perhaps but I'll let @iantaylor-NOAA chime in here. R doesn't love spaces in text or special characters other than "_" or "." so I would avoid those 2 things when it comes to easy compatibility with r4ss. 

Is Tot_Exploit under a different column header or is it no longer reported? It's used in r4ss in the get_SIS_info() which is why I am asking.

Rick-Methot-NOAA: Tot_Exploit got renamed to:  dead_catch_B/bio_smry.
I can switch back to Tot_Exploit and put dead_catch_B/bio_smry into a note.

Delimiters:  I already was using _ and will continue to do so.

Probably useful to look through the report file and describe all the different formats for notes that exist already.  I am especially interested in making it easy for r4ss to know that a line of text is a string of headers.

e-perl-NOAA: Okay, great. It's fine the way it is, I just was trying to figure out what it had become.

iantaylor-NOAA (approved):
Looks good. I think having a separate table of annual quantities will be really useful for assessment reports in the future.

One question: the columns labeled `SPR_std` and `F_std` in the new table were called `SPR_report` and `F_report` in the old table. It looks like the new names match the code, but if there's not a compelling reason for the change, I would switch back to the old names which I think are better aligned with "SPR_report_basis" and "F_report_basis" as well as requiring fewer changes in the r4ss code.

Rick-Methot-NOAA:
@iantaylor-NOAA - regarding F_report vs. F_std.  They refer to the same quantity and both terms are used in various places through the report file, with more for F_std.  I will consolidate all usage to F_std which is the name of the array.  The term's history is that it is the F quantity for which the standard deviation (std) is produced and reported.

iantaylor-NOAA:
Makes sense, thank you. Easy enough to adapt.

Issue #611, opened 2024-08-02
revise RECR_DIST output to work better for models without 50/50 sex ratio at recruitment?


iantaylor-NOAA:
The 2nd dimension of the `recr_dist` array has length equal to the product of the number of growth patterns and number of sexes:
https://github.com/nmfs-ost/ss3-source-code/blob/69cc71611cf790d520a53f7ee47fe6885fdeb8fe/SS_param.tpl#L157
However the RECRUITMENT_DIST output in Report.sso only loops over growth patterns, not sexes within them:
https://github.com/nmfs-ost/ss3-source-code/blob/69cc71611cf790d520a53f7ee47fe6885fdeb8fe/SS_write_report.tpl#L1031-L1043

This works fine for the huge majority of models that are either 1-sex or have the fraction female fixed at 0.5.

However, @akatan999 shared a model (shared with SS3 developers on Google Drive [here](https://drive.google.com/drive/folders/1Uy8yju7HlLt6m8SrMGwAw_k9k2W-NprB)) with hermaphroditism and multiple areas where the fish recruit to both areas as males but the values in the table only show the tiny fraction of females recruiting to each area:
```
RECRUITMENT_DIST report:10
Settle# settle_timing# G_pattern Area Settle_Month Seas Age Time_w/in_seas Frac/sex
1 1 1 1 1 1 0 0 4.43463e-07
2 1 1 2 1 1 0 0 5.56537e-07
```

Could that table be revised to either sum across females and males OR have separate rows for both females and males?

iantaylor-NOAA:
Update: I just changed the code in https://github.com/r4ss/r4ss/commit/2081ce2df76e12358a50bbbf16c4ab339b5329e8 to rescale the recruitments to sum to 1.0 before plotting. It may still be worth revising the SS3 output but the confusing plot which led to this issue has been resolved:
![image](https://github.com/user-attachments/assets/f81b5091-9125-414c-ba14-f3b02f236c31)

Rick-Methot-NOAA:
The values being plotted are just the frac female.  I will revise the SS3 output table to also output the fraction male.

PR #612, opened 2024-08-05
revise recr_dist output to show both sexes

Rick-Methot-NOAA:
revise recruitment distribution output to show males and females.  This is helpful for hermaphroditism applications

* Resolves issue #611 

## What tests have been done? 
see test file linked in the issue.
revised output is visible in the issue and shows the desired feature

### DIFF
--- a/.github/workflows/test-r4ss-with-ss3.yml
+++ b/.github/workflows/test-r4ss-with-ss3.yml
@@ -50,7 +50,7 @@ jobs:
         run: Rscript -e 'install.packages(c("remotes","parallely", "furrr", "future"))'
 
       - name: Install r4ss
-        run: Rscript -e 'remotes::install_github("r4ss/r4ss", ref = "try-fix-matrix-error")'
+        run: Rscript -e 'remotes::install_github("r4ss/r4ss")'
 
       # - name: Get admb and put in path, linux
       #   run: |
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA (approved):
Looks good. Thanks for the quick fix @Rick-Methot-NOAA. 

In hindsight it's easy enough to compute the male distribution as 1 - females, but I think having all the output there makes things more transparent and easier to access.

I'll finish modifying r4ss to work with this revised output tomorrow (mostly just requires looping over sexes before making the figure).

iantaylor-NOAA:
The test-r4ss-with-ss3 github action is failing [here](https://github.com/nmfs-ost/ss3-source-code/actions/runs/10256837962/job/28472647692?pr=612#step:12:216) in modifications I made to the `SSplotRecrdist()` plot to take advantage of the output for males and females. However, I can't replicate the issue. The two models that are causing problems, Simple_with_DM_sizefreq and two_morph_seas_areas, work fine when running with the SS3 exe from this 611... branch and the main branch of r4ss on my local computer. Trying to trap for incorrectly specified values in the r4ss function didn't work either and it's hard to debug without replicating the problem.

I tried running a github codespace to debug further but am having trouble getting that running. I will be in Sea Safety Training for the next few days but can try again on Monday. Hopefully the delay isn't a big deal as I don't think the changes in this PR interact with other work.

@e-perl-NOAA if you feel like trying to debug via codespaces or any other method before I get back to it, feel free.

e-perl-NOAA:
The r4ss issue is resolved so this should be good to go now.

Issue #615, opened 2024-08-15
[Bug]: rho parameters not impacting semi-parametric (2D-AR) selectivity when multiple sigma values are used


iantaylor-NOAA:
### Describe the bug

This issue was found in discussion with @rosana-ourens who successfully set up a model with semi-parametric selectivity where there were multiple sigma values to allow more flexible deviations for the youngest ages. However, the rho parameters did not seem to have any impact on the results.

I replicated the issue in the attached models, modified from https://github.com/nmfs-ost/ss3-test-models/tree/main/models/Hake_2019_semi-parametric_selex, which have a similar set up where sigma_amax = 3 and rho_year and rho_AGE are both fixed at either 0 or 0.5:
[sigma_3_rho_0.zip](https://github.com/user-attachments/files/16631204/sigma_3_rho_0.zip)
[sigma_3_rho_0.5.zip](https://github.com/user-attachments/files/16631206/sigma_3_rho_0.5.zip)

The likelihoods differ between models but the difference is all due to the Parm_devs component as shown in the comparison below.
![image](https://github.com/user-attachments/assets/7e973166-96b3-4b99-ad58-8f84598672f8)

Changing sigma_amax from 3 to 0 and reducing from 3 to 1 sigma_sel parameters, allows the rho parameter to have a modest impact on the selectivity parameters and all other aspects of the dynamics.

This may be due to my misunderstanding of how the rho parameters work, but I was assuming that they apply across all ages and years of deviations regardless of the how the sigmas are configured.

### To Reproduce

Run the models attached above.

### Expected behavior

Different rho values impact the model results beyond just the parameter deviations.

### Which version of SS3 are you seeing the problem on?

3.30.22.1

Rick-Methot-NOAA:
I see in echoinput that the determinant for 2D_AR cor: 1  is: 1.839e-30

So, we may need to consult @Haikun  on this.

The code is here:  https://github.com/nmfs-ost/ss3-source-code/blob/f2df4f49c177e718ad4ac8aa13aabe4adb26102a/SS_prelim.tpl#L1020

iantaylor-NOAA:
@HaikunXu, let us know if you have time to look at the code in https://github.com/nmfs-ost/ss3-source-code/blob/3191eb82b6aabb27abd4e500744af4b69cc2b821/SS_prelim.tpl#L1020 and let us know if you see anything wrong related to the implementation of rho parameters when there are multiple sigma values?
[...]

PR #617, opened 2024-09-13
adjust warnings for 2DAR

Rick-Methot-NOAA:

Find that the rho feature of 2DAR was incompletely implemented and a full fix is not feasible at this time.  Provide new warnings and guidance to users.

* Resolves issue #615 

## What tests have been done? 
Numerous runs with various permutations of estimating or fixing age-specific sigmasel values using hake 2DAR example.
none

### DIFF
--- a/SS_readcontrol_330.tpl
+++ b/SS_readcontrol_330.tpl
@@ -4521,20 +4521,20 @@
   TwoD_AR_cnt = 0;
   echoinput << " now read 0/1 for 2D_AR" << endl;
   *(ad_comm::global_datafile) >> TwoD_AR_do;
-  echoinput << TwoD_AR_do << "  #_ 0/1 to request experimental 2D_AR selectivity smoother options " << endl;
+  echoinput << TwoD_AR_do << "  #_ 0/1 to request experimental 2D_AR selectivity deviations " << endl;
   
   if (TwoD_AR_do > 0)
   {
-    warnstream << "2D_AR selectivity deviations option is selected!";
       //  elements 1-11 are read from control.ss;  12 and 13 are calculated internally
       //  1-fleet, 2-ymin, 3-ymax, 4-amin, 5-amax, 6-sigma_amax, 7-use_rho, 8-age/len, 9-dev_phase
       //  10-before yr range, 11=after yr range, 12-N_parm_dev,  13-selparm_location
-    write_message (WARN, 0);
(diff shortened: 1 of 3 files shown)


SS_selex.tpl
```diff
@@ -1740,8 +1740,8 @@ FUNCTION void get_selectivity()
                 if (k >= 0)
                 {
                   k++;
-                  if (docheckup == 1)
-                    echoinput << a << "  sigmasel: " << sigmasel << endl;
+//                  if (docheckup == 1)
+                  temp = sel_a(y, fs, 1, a);
```
iantaylor-NOAA, line 1744: I don't understand the impact of this `temp` variable

iantaylor-NOAA (approved):
Looks good to me. Thanks for all the explorations.

SS_selex.tpl
Rick-Methot-NOAA: Oops.  That's a leftover from my testing.  I'll clean it out.

Rick-Methot-NOAA:
@e-perl-NOAA   Please reflect these revised warnings into the manual.

SS_readcontrol_330.tpl
```diff
@@ -4604,14 +4604,22 @@
...
+          if (dtempvec(7) > 0)
+          {
+            warnstream << "2DAR Sigmasel parameter is not normally estimable; adequate performance usually obtained by fixing to value of 1.0";
+            write_message (WARN, 0);
+          }
         }
         if (use_rho == 1)
         {
-          echoinput << "read two parameter lines for rho_yr and then rho_age (or length)" << endl;
+          warnstream << "2DAR rho is incompletely implemented; it should only be used experimentally and never estimated";
```
kellijohnson-NOAA, line 4615: I am hesitant to allow users to even do this if it is incompletely implemented. I would prefer a full stop if `use_rho = 1`. Especially because in this same if statement you set `dtempvec(7) = -1`. Which, does that just change the values put in ss_new or does it actually turn off estimation without the user knowing?

kellijohnson-NOAA (commented):
I am uncertain of the level of guidance that you should be giving the user. It seems like you are just warning them and then changing it behind the scenes, where I prefer that the model just stops if they try to estimate it.

SS_readcontrol_330.tpl
Rick-Methot-NOAA: The estimation has never been enabled.  The value of rho is only used in the preliminary calcs section to create a cor matrix.

Issue #619, opened 2024-09-18
[Bug]: timing for spawnbio calculation incorrect in seasonal model and spawning not at start of season


Rick-Methot-NOAA:
### Describe the bug

In a seasonal model with spawnmonth not set to the beginning of a season, the reported spawnbio was lower than expected.
Investigation found that there was no discrepancy with spawnmonth at beginning of a season.
And that the magnitude of difference was in accord with being off by exp(-z*seasdur)

The model output for spawntiming as shown in echoinput was:
2 N seasons 
 6 6 months/seas (fractions OK) 
2 Number of subseasons (even number only; min 2) for calculation of ALK 
 0.5 0.5 processed season duration (frac. of year) 
 0.25 0.25 processed subseason duration (frac. of year) 
 processed subseason cumulative annual time within season 
 0 0.25
 0 0.25
SPAWN month: 5.5; seas: 1; subseas_for_ALK: 2; timing_in_season: 0.75

Note that timing is as a fraction of the season.  When the value gets used in internal calculations, season duration was not being taken into account:
`SSB_pop_gp(y, p, GP4(g)) += fracfemale_mult * fec(g) * elem_prod(natage(t, p, g), mfexp(-Z_rate(t, p, g) * spawn_time_seas)); // accumulates SSB by area and by growthpattern`

Solution will be to adjust spawn_time_seas by duration of the season.

### To Reproduce

per what is shown above

### Expected behavior

spawnbio calculation to decay by exp(-Z) within the season

### Which version of SS3 are you seeing the problem on?

This has probably been an error throughout the history of 3.30.

PR #620, opened 2024-09-18
revise calc of spawn_time_season

Rick-Methot-NOAA:

* Resolves issue #619 

## What tests have been done? 
season duration calculation have been compared to season duration for a survey with same timing.  They did not match before, now they do.

Test file is in email with nmfs.stock.synthesis

None

### DIFF
--- a/SS_readdata_330.tpl
+++ b/SS_readdata_330.tpl
@@ -199,9 +199,10 @@
       }
       temp += subseasdur_delta(spawn_seas);
     }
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA (approved):
Looks good. Nice job with the quick turnaround.

Rick-Methot-NOAA:
The nice work was by Leire Citores (AZTI) who did the investigation showing that the timing calculation was wrong.

Rick-Methot-NOAA:
note that nothing in our test suite was affected by this error.  It is just the unusual cases with multi-seasons and spawn_month not at beginning of a season.

PR #621, opened 2024-09-19
Update build-ss3.yml with admb build from source for macos-13

e-perl-NOAA:

The macos-12 build using the admb docker image started failing after an update to the macos-12 github action runner. The macos-13 runner was also failing to build ss3 with the admb docker image. The workflow has been updated to use the macos-13 runner and build admb from source for that runner and then build ss3 rather than using the admb docker image.

See the [successful build](https://github.com/e-perl-NOAA/github-actions-test/actions/runs/10947197746) in my repo with the workflow.

None.

### DIFF
--- a/.github/workflows/build-ss3.yml
+++ b/.github/workflows/build-ss3.yml
@@ -10,8 +10,10 @@ on:
       - '**.sh'
     workflow_dispatch:
 
+
 jobs:
   build-ss3:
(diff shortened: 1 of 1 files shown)


Rick-Methot-NOAA (approved):
Do you think we can update the Docker later?

e-perl-NOAA:
Possibly but the problem is that I'm not getting any (or any useful) error messages from the GitHub action, it just gets canceled and doesn't finish running so I'm not sure what the problem is or how to fix it. I'm hoping that maybe it's a bug with the runner that will get worked out in time but I'm really not sure.

Issue #486, opened 2023-07-27
[Refactor]:  make use of join functions more consistent


Rick-Methot-NOAA:
### Refactor request

A user noticed that a forecast result did not match expectation output (see issue #485 ).  This is due to the control rule's JOIN not being steep enough, which allows values >1.0 to occur near the transition

The logistic join function is used to smoothly connect two functions without using an IF statement.  It is used in double normal selectivity, control rule inflection, hockey stick SRR, and other places.  
If the join is not steep enough, then the resultant value of the join can be > 1.0 close to the transition point.

The function shown below is not used universally to achieve the join.  It has a very steep transition (1000).  Other custom join implementations use  10, 20, 30, 40, 50, 100.  10 is the value in the control rule join:
```
FUNCTION dvariable Join_Fxn(const prevariable& MinPoss, const prevariable& MaxPoss, const prevariable& Inflec, const prevariable& Xvar, const prevariable& Y1, const prevariable& Y2)
  {
  RETURN_ARRAYS_INCREMENT();
  dvariable Yresult;
  dvariable join;
  join = 1.000 / (1.000 + mfexp(1000.0 * (Xvar - Inflec) / (MaxPoss - MinPoss))); //  steep joiner at the inflection
  Yresult = Y1 * (join) + Y2 * (1.000 - join);
  RETURN_ARRAYS_DECREMENT();
  return Yresult;
  }

```
This should be cleaned up after an evaluation of a good default steepness (suggest 100).

- [ ] create test case to guide selection of optimal steepness for the joiner
- [ ] rewrite all joiner uses to use the function:  Join_Fxn

### Expected behavior

more consistent code

Rick-Methot-NOAA:
@iantaylor-NOAA  - here is the issue regarding joiners.

Rick-Methot-NOAA:
`C:\Users\Richard.Methot\Documents\GitHub\StockSynthesis_git\stock-synthesis\SS_benchfore.tpl
  2001,5:     join1 = 1. / (1. + mfexp(30. * (Fcast_Fmult - max_harvest_rate)));
  2632,13:             join1 = 1. / (1. + mfexp(10. * (SSB_current - H4010_bot * temp)));
  2831,23:                       join1 = 1. / (1. + mfexp(30. * (temp1 - max_harvest_rate)));
  2857,23:                       join1 = 1. / (1. + mfexp(30. * (temp1 - max_harvest_rate)));
[...]

Issue #485, opened 2023-07-27
[Bug]: ForeCatch not quite equal to OFLCatch * buffer


iantaylor-NOAA:
### Describe the bug

The attached petrale sole model [2023.a034.011_forecast_SR.zip](https://github.com/nmfs-stock-synthesis/stock-synthesis/files/12177669/2023.a034.011_forecast_SR.zip) (input and output) has a forecast buffer that is set to 0.935 in 2025. 

The OFLCatch_2025 is 2517.67 and the ForeCatch_2025 is 2356.47, but applying the 0.935 buffer to the OFL is different by about 2 tons: 2354.021. In that year the biomass is slightly above the 0.25 control rule inflection for flatfish so no ramping down in forecast catch should be occurring (unlike subsequent years). I see that the value of ABC_buffer in TIME_SERIES is 1.00119 for 2025, so maybe that's the source of the difference, but I'm not sure how that value is calculated or if I have something set wrong in the forecast file.

The differences are small, but large enough to be noticed by the review panel. And one other model that I looked at had an exact match (within the 0.01 precision reported).

@Rick-Methot-NOAA, do you think this is just some form of rounding error or the result of an iterative process that doesn't provide an exact match, or is there something else going on? 

And if there's not a quick solution, do you have guidance on what would be the most accurate source of ACL values for management: the ForeCatch from the model or the OFLCatch multiplied by the buffer outside the model?

### To Reproduce

Look at attached files.

### Expected behavior

more precise equality for ForeCatch_2025 = OFLCatch_2025 * 0.935

Rick-Methot-NOAA:
The logistic joiner for the control rule is almost certainly the source of this small discrepancy.
https://vscode.dev/github/nmfs-stock-synthesis/stock-synthesis/blob/main/SS_benchfore.tpl#L2624-L2648

If you need exact, applying buffer to OFL seems logical alternative

chantelwetzel-noaa:
I have looked into this issue as well. I am not convinced that there is an issue in the joiner for the harvest control rule since this issue is not occurring for two different models (canary and copper). The canary model is using the same executable as Ian's model (#V3.30.21.00;_fast(opt);_compile_date:_Feb 10 2023) and the copper model is also using the same version but the safe executable. However, looking over the model files that Ian has shared I have been unable to find a setting that would be causing the buffer value to not be applied correctly.

Rick-Methot-NOAA:
[...]

PR #623, opened 2024-09-24
first commit for joinsteep=50

Rick-Methot-NOAA:

Replace all the individual join function constants with the constant termed joinsteep and with a value of 50.  Values previously ranged from 10 to 1000.

* Resolves issue #486 

## What tests have been done? 
see comparisons in the issue

### DIFF
--- a/SS_benchfore.tpl
+++ b/SS_benchfore.tpl
@@ -1998,7 +1998,7 @@ FUNCTION void Get_Forecast()
         break;
       }
     }
-    join1 = 1. / (1. + mfexp(30. * (Fcast_Fmult - max_harvest_rate)));
+    join1 = 1. / (1. + mfexp(joinsteep * (Fcast_Fmult - max_harvest_rate)));
     Fcast_Fmult = join1 * Fcast_Fmult + (1. - join1) * max_harvest_rate; // new F value for this fleet, constrained by max_harvest_rate
     if (join1 < 0.999)
     {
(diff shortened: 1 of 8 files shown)


Rick-Methot-NOAA:
perhaps I should revert back to the previous big values (1000) and small values (10) to see which had the biggest impact.

iantaylor-NOAA (approved):
The changes all look good, but I'm having second thoughts about the value of making all the join functions the same. I think there are at least 3 distinct cases that might benefit from separate values: 
1. join functions used for piecewise linear functions like the harvest control rules and the hockey stick stock-recruit function. Ideally the transitions would be steep so that the resulting patterns are as linear as possible and you don't get mismatches from the expectation as discussed in #485 and #486. Changing the values should only impact models where the population crosses one of the joins in a way that impacts the likelihood. Thus, changes to the join in the HCR shouldn't impact estimation because it's only applied in the forecast.
2. join functions used for selectivity, e.g. smooth transition from ascending limb to flat top of double normal. I don't see any clear benefit from having those joins be steeper as we would expect the selectivity curves to be pretty smooth. Furthermore, making a change here will likely have a small impact on the large majority of models which use double normal selectivity.
3. join functions that relate to hitting limits like max F or crash penalty. I don't know that it matters how steep these are. Too steep may hinder estimation for those models that hit such bounds, but the MLE values may not be impacted as long as the estimation leaves the parameter space which caused the problem (though the different search path may lead to slightly different values).

Issue #622, opened 2024-09-23
[Bug]: Unable to include multiple CAAL records for same fleet, year, and bin with different ageing errors


chantelwetzel-noaa:
### Describe the bug

I was running some exploratory runs to include ages from a new ageing method and ran into a strange error.  The model was including two sets of CAAL data from the same fleet and years using two separate ageing errors.  The message in the terminal was fairly cryptic:

Warning 1 Fatal Error! fatal:  max agecomp obs per fleet*time is 100; you requested 101 for fleet x year 7 2017

When I removed one set of the CAAL data, the model ran.  I then tried running the model with the two separate CAAL datasets but had them use the same ageing error and I got the same fatal error message.  Given that, I think the issue arises when a user in inputting more than one line of CAAL data for the same year, fleet, and length bin.  

### To Reproduce

I will email the model files directly where there are two sets of CAAL data from 2017-2022, one labeled traditional-ages and one called ft-nirs.  

### Expected behavior

If this is a relatively easy fix, allowing users to include CAAL from multiple age reading methods would be useful. Next year's West Coast sablefish assessment may have this need, and allowing the use of all data will allow us to include all data as CAAL.

### Which OS are you seeing the problem on?

Windows

### Which version of SS3 are you seeing the problem on?

3.30.21

Rick-Methot-NOAA:
got it.  This affects indexing of observations, so may be a big deal to fix.  I'll take a look.

PR #624, opened 2024-09-25
increase have_data storage to 150.

Rick-Methot-NOAA:

simple increase in storage of have_data array to allow more CAAL observations.  New limit is 150.

* Resolves issue #622 

## What tests have been done? 
revised sablefish model runs with this increased storage
none

 [x] No, there was no input change. -->

```
If there is no issue related to this PR, replace this text with your example stock synthesis input.
```
-->


chantelwetzel-noaa (approved):
Thank you for fixing this so quickly.  Everything looks good to me.

PR #641, opened 2024-11-07
Build macos-latest admb from source

e-perl-NOAA:

Per the usual, mac is always the problem child. While the macos-latest exe was building using the admb docker image, it wasn't actually working which wasn't detectable until r4ss r_cmd_check() tests ran on macos-latest. Using the build from source workflow, I created a ss3_osx_arm64 exe that works, so I have modified our ss3-build workflow to build both versions of mac from source.

In the future, I think  running the test models with macos latest in addition to it running on linux would be useful, especially since no one has a M2 chip and beyond mac to test on anymore (my mac is old). It would increase the run time but it would ensure that we aren't finding out that it doesn't work later than we want. I'm happy to work on this if you agree. I can even add it as a part of this PR. It won't take long to implement.

None.

### DIFF
--- a/.github/workflows/build-ss3.yml
+++ b/.github/workflows/build-ss3.yml
@@ -58,19 +58,19 @@ jobs:
           
       # Build ADMB for macos-13
       - name: Build ADMB for macos-13 and put in path
-        if: matrix.config.os == 'macos-13'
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA (approved):
Looks good, thanks for dealing with this @e-perl-NOAA.

Issue #646, opened 2024-11-29
[Bug]: model crash during output stage when using Qpower


Rick-Methot-NOAA:
### Describe the bug

When Qtype = 3 (power function) is used, the model crashes during output calculations.
Finished calculating benchmarks
Finished forecast
Writing wtatage.ss_new
mceval counter: 0
Finished posteriors
terminate called after throwing an instance of 'std::out_of_range'
  what():  Invalid index in param_init_number_vector::operator()(int i)

DIAGNOSIS:
When parameter values for offset and power were added to the output table Index_1, the logic for what to report was developed using a test case with both offset and power (Qtype = 6).  That logic broke for Qtype = 3.

While fixing this logic problem, I aslo added an output note under Index_2:
NOTE: Calc_Q is Q_base with devs, so does not include power; eff_Q includes effect of Power

Also removed two extraneous output statements in SPR/YPR profile

### To Reproduce

none

### Expected behavior

proper result

### Which version of SS3 are you seeing the problem on?

3.30.22 and current

Rick-Methot-NOAA:
[test_Qpower.zip](https://github.com/user-attachments/files/17963167/test_Qpower.zip)

PR #647, opened 2024-11-29
issue 646 - fix output error in Index_2, plus two minor output tweaks

Rick-Methot-NOAA:

fixes output error in Index_2 that was introduced with Qtype 6.

* Resolves issue #646 

 [x] Test files are in the issue. 

None

### DIFF
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -1912,6 +1912,7 @@ FUNCTION void write_bigoutput()
     {
       SS2out << endl
              << pick_report_name(22) << endl;
+      SS2out << "NOTE: Calc_Q is Q_base with devs, so does not include power; eff_Q includes effect of Power" << endl;
       //  where show_time(t) contains:  yr, seas
       //  data_time(ALK,f) has real month; 2nd is timing within season; 3rd is year.fraction
       //  show_time2(ALK) has yr, seas, subseas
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA (approved):
The changes look good to me. Thanks for fixing the issue so quickly.

The "Q" to "Qbase" change wasn't the issue for r4ss, rather the logic for parsing the INDEX_2 table wasn't adaptable to the additional comment line (unlike most other tables). I think I fixed the r4ss issue on the second try and just restarted the failed github action for the second time. If it passes the check (should be done in about 15 minutes) I think it's fine to merge.

iantaylor-NOAA:
The `test-r4ss-with-ss3` check is now passing, so I think this is fine to merge.

@e-perl-NOAA do you know why `build-ss3` was skipped?

@Rick-Methot-NOAA, do you think this bug needs a bug-fix release? It seems like users interested in Q_type 3 = power (+1 parm) which is causing the error fixed by this PR. could temporarily use type 6=offset & power (+2p) with the offset parameter fixed at 0 and get all the functionality of type 3. If we announced that solution on the forum we could potentially avoid a bug fix release and just wait until other features are ready for a bigger release.

Rick-Methot-NOAA:
I thought the same Ian.  Let's do a bug fix release for this one.

e-perl-NOAA:
I have no idea why it was skipped, but I'm running it [here](https://github.com/nmfs-ost/ss3-source-code/actions/runs/12144400962)

Rick-Methot-NOAA:
Let's do both a big fix and a message in the Forum.  The Forum message can serve to advertise the availability of the offset approach.

iantaylor-NOAA:
Sounds good.
A bug fix release has the additional benefit that we can test whether the linux and mac versions continue to have issues with missing libraries when run in those same platforms on github actions (necessitating an extra setup step to add the libraries as @e-perl-NOAA did at https://github.com/nmfs-ost/ss3-user-examples/commit/7173d5e83ffde4f3c9d04e44c6cf91bd3067cfd1).

e-perl-NOAA:
Sounds good!

Issue #662, opened 2025-03-31
[Bug]: age discard with discard mortality using wrong vector


Rick-Methot-NOAA:
### Describe the bug

The problem is that the vector created to represent dead fish not using a correct component.

It is calculated here: 
`               discmort2_a(y, fs, 1) = elem_prod(sel_a(y, fs, 1), retain_a(y, fs, 1) + elem_prod((1. - retain_a(y, fs, 1)), discmort_a(y, fs, 1)));`

then with docheckup turned on you can get it output as: 
`         if (docheckup == 1 && y == styr)            echoinput << " dead " << discmort2_a(y, fs) << endl;
`
But I do not see it getting used in a place like here: 
`           sel_dead_bio(s, f, g) = elem_prod(sel_ret_bio(s, f, g), discmort_a(y, f, gg)); //  dead wt            sel_dead_num(s, f, g) = elem_prod(sel_ret_num(s, f, g), discmort_a(y, f, gg)); //  dead numbers.`

The problem seems related to which vectors get created and stored with a "y" subscript, and which are created from them and stored with a "s" season subscript for use in the catch calculations.

### To Reproduce

reported dead catch in a model with age-based retention is shown as discard_mortality * retained_catch, rather than the correct value

### Expected behavior

 //  4darray sel_dead_num(1,nseas,1,gmorph,1,Nfleet,0,nages);  // sel * (retain + (1-retain)*discmort)
  //  4darray sel_dead_bio(1,nseas,1,gmorph,1,Nfleet,0,nages);  // sel * (retain + (1-retain)*discmort) * wt

### Which version of SS3 are you seeing the problem on?

3.30.22.  Probably has been incorrect since age-retention was created

PR #663, opened 2025-04-01
age-based calc of sel_dead_num and bio is repaired

Rick-Methot-NOAA:

* Resolves issue #662

## What tests have been done? 
when use age-based selectivity, retention and discard mortality, the results now appear correct with the sel_dead values intermediate between the sel and the sel*ret values

### DIFF
--- a/SS_selex.tpl
+++ b/SS_selex.tpl
@@ -1973,10 +1973,11 @@ FUNCTION void Make_FishSelex()
           }
           case 2:
           {
+            // details of retention and mortality already taken into account with calc of discmort2_a
             sel_ret_bio(s, f, g) = elem_prod(sel_bio(s, f, g), retain_a(y, f, gg)); //  retained wt-at-age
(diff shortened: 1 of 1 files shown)


SS_selex.tpl
```diff
@@ -1973,10 +1973,11 @@ FUNCTION void Make_FishSelex()
           }
           case 2:
```
e-perl-NOAA, line 1974: @Rick-Methot-NOAA Is it worth adding a comment to case 1 and case 2 like you have with case 3 where it says 'all selected fish are dead'

iantaylor-NOAA (approved):
The code changes look good to me, but I would leave it to @chantelwetzel-noaa to confirm that the results for her model look accurate.

Rick-Methot-NOAA:
good suggestion Elizabeth to clean up the code comments, some of which were copied between sections.
BTW, I made a comment in another issue about what looks like a non-trivial bit of work to get separate wt-at-age for discarded fish into the wtatage..ss file and used by the code.

chantelwetzel-noaa (approved):
I tested the model and the values in the catch table are as expected.

PR #667, opened 2025-04-07
Update build-ss3.yml - remove if statement

e-perl-NOAA:
if statement on build is causing the workflow to be confused, best to just remove

There was an if statement that was causing the workflow to get confused and occasionally not run. Removed statement so that it runs.

None.

### DIFF
--- a/.github/workflows/build-ss3.yml
+++ b/.github/workflows/build-ss3.yml
@@ -13,7 +13,6 @@ on:
 
 jobs:
   build-ss3:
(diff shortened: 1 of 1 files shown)


e-perl-NOAA:
I have access to copilot as a part of the pilot project, I'm testing using copilot as a reviewer of this PR too just out of curiosity.

Rick-Methot-NOAA (approved):

iantaylor-NOAA (approved):
I've been confused about when/why this runs, so thanks for cleaning it up. It's easy to delete the extra notifications for the runs that we don't care about.

Issue #670, opened 2025-04-09
[Bug]: indexing error in age-selex 41


Rick-Methot-NOAA:
### Describe the bug

parameter referencing showed corruption when age-selex 41 was attempted.

[age_selex_41.txt](https://github.com/user-attachments/files/19671066/age_selex_41.txt)

### To Reproduce

selectivity output was wrong

### Expected behavior

logical selectivity pattern

PR #671, opened 2025-04-09
#670 repair bug in selex 41

Rick-Methot-NOAA:

repairs an indexing error for parameters in age selex 41 that scales over a defined age range

* Resolves issue #670

### DIFF
--- a/SS_readcontrol_330.tpl
+++ b/SS_readcontrol_330.tpl
@@ -3721,6 +3721,8 @@
     else
     {
       N_selparmvec(f) = abs(seltype(f, 4)) + 1; // so reads value for age 0 through this age
+      if (seltype(f, 1) == 41) N_selparmvec(f) +=2;  //  for the scaling parameters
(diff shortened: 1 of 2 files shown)


e-perl-NOAA:
@Rick-Methot-NOAA Did you already confirm that it works on Jim's files?

jimianelli:
thanks!

jimianelli:
Seems to work as advertised. Keep track of which ages the parameters affect.

e-perl-NOAA:
@jimianelli, Is there any additional guidance in the user manual that needs to happen with this?

jimianelli:
great question. In the section describing selex 41 I would be explicit about the order and meaning of the parameter specifications. need to define that nages = max age (typically excludes zero...). Off the top it goes:
p1: low-age
p2: high-age (where these are the age range over which to compute mean for normalizing)
then the description in selex 17. So in the table:
image
put a "...." after 2nd row and something like "As in selex 17" so the order is clearer, and maybe bold face the last sentence before this table.

Issue #678, opened 2025-05-23
[Bug]: control.ss_new not writing 2D-AR1 parameter lines when sigma_amax < 0


iantaylor-NOAA:
### Describe the bug

The User Manual [guidance on 2D-AR1 (semi-parametric) selectivity](https://nmfs-ost.github.io/ss3-doc/SS330_User_Manual_release.html#TwoDAR) notes that "For simplicity, only a single sigma parameter line needs to be read if Sigma_Amax equals Amin, or if Sigma_Amax < 0."

However, the code to write the control.ss_new file doesn't include this option and just skips writing of the one parameter line: https://github.com/nmfs-ost/ss3-source-code/blob/b8a5b59b67093e8346dce47b5c3a0e9ada541df2/SS_write_ssnew.tpl#L2627-L2634.

The example attached includes input files (which runs successfully) and the control.ss_new file with the missing parameter line.
[simple_small_2DAR_len.zip](https://github.com/user-attachments/files/20415739/simple_small_2DAR_len.zip).

### To Reproduce

Run the model attached above.

### Expected behavior

Parameter line for 2D-AR selectivity

### Which version of SS3 are you seeing the problem on?

3.30.23.1

iantaylor-NOAA:
I forgot to credit @EJDick-NOAA for sharing a model that includes the setup replicated in the attached example where the r4ss function for reading the control file also fails to work with sigma_amax < 0.

Rick-Methot-NOAA:
assignment accepted

Rick-Methot-NOAA:
revised code produces in control.ss_new:
```
#_specs:  fleet, ymin, ymax, amin, amax, sigma_amax, use_rho, len1/age2, devphase, before_range, after_range
#_sigma_amax>amin means create sigma parm for each bin from min to sigma_amax; sigma_amax<0 means just one sigma parm is read and used for all bins
 3 2011 2022 5 15 -1 0 1 6 0 0  #  2D_AR specs for fleet: SURVEY2 LEN
 0.1 2 1 1 99 0 -3  # sigma_sel for fleet:_3; LEN_5
-9999  0 0 0 0 0 0 0 0 0 0 # terminator
```

iantaylor-NOAA:
Looks good, thanks for working on this so quickly @Rick-Methot-NOAA

PR #679, opened 2025-05-23
fix 2dar control_ssnew

Rick-Methot-NOAA:

add code to allow 2dar output to control.ssnew when sigma_amax<0

* Resolves issue #678

## What tests have been done? 
verified that revised code produces the expected line in control.ssnew

<-- - [x] Test files are in the issue. -->

None

## Additional information (optional).

@EJDick-NOAA


iantaylor-NOAA (approved):
Looks good. This fix is simpler than I expected.

PR #640, opened 2024-11-06
191 spawner recruit rebased

Rick-Methot-NOAA:
continuation of PR #442 after updating from 3.30.23 main on 11/06/2024
This implements SRR function 10 for Bev-Holt with alpha, beta
also adds two new controls to give user more control of how time-varying biology affects benchmark and control rule implementation.

Resolves issue #191, #341, #625

See the checkboxes at top of issue #191 for the list of major changes with this branch.

PR #683, opened 2025-06-26
minor refinements to headers in SPR and YPR output

iantaylor-NOAA:
Changes `SSBpR` to `SPR` and `YpR` to `YPR` in table headers and comments. This causes the terminology used in the output differ from variable names in the TPL, but it makes the output internally consistent and more consistent with previous versions of SS3 and common terminology, which will also simplify processing by r4ss. Suggesting these as a Pull Request to the in-process branch seemed easier than describing the changes in a comment on the existing PR #640.

The one change I was least confident in is `SPR_virgin`. It seems like it could be either `SPR0` or `SPR_virgin`, but `SPR0_virgin` seems redundant.

Changes are only to comments in Report.sso so should not impact anything numeric. 

@Rick-Methot-NOAA if you think any of these changes aren't good, feel free to close this PR or change things around.

### DIFF
--- a/SS_write_report.tpl
+++ b/SS_write_report.tpl
@@ -1919,8 +1919,8 @@ FUNCTION void write_bigoutput()
     { if (Bmark_Yr(k+1) > Bmark_Yr(k))
       {SS2out << "#_NOTE:_using_range_of_years_can_reduce_standard_error_of_result;_do_this_only_when_timevarying_makes_necessary:  " << k << " " << Bmark_Yr(k) << " " << Bmark_Yr(k+1) << endl;}
     }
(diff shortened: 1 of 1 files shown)


Rick-Methot-NOAA:
I will take a look.  My intention with the extensive refactoring was to clarify that:
SPR = SSBpR(fished) / SSBpR(unfished)
but I may have missed some.
The original code often used SPR when the quantity was actually a SSBpR

Rick-Methot-NOAA (commented):
capitalization in line 4795 is good change.
line 4797 with word "stored" can be safely deleted as it looks like a Rick testing line
other changes should not be made

Issue #685, opened 2025-07-05
[Feature]: Portable Linux executable


arni-magnusson:
### Describe the solution you would like.

Hi! I downloaded the latest Stock Synthesis for Linux on the release page ([v3.30.23.2](https://github.com/nmfs-ost/ss3-source-code/releases/tag/v3.30.23.2), ss3_linux).

When I try to run a model, I get the following error message:
```
ss3_linux: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by ss3_linux)
```

Can I suggest that the Linux executables made available on the release page be built with the `-p` option? This would make `ss3_linux` portable, so it runs on all Linux machines.

I'm guessing the relevant code is in one of the GitHub workflows, either [build-admb-and-ss3-from-source.yml](https://github.com/nmfs-ost/ss3-source-code/blob/main/.github/workflows/build-admb-and-ss3-from-source.yml) or [build-ss3.yml](https://github.com/nmfs-ost/ss3-source-code/blob/main/.github/workflows/build-ss3.yml). These workflows call the `Make_SS_330_new.sh script` multiple times, sometimes with `-p` and sometimes without. For Linux users, it would be better if the Linux executables made available on the release page would be the portable ones.

P.S. My specific use case is a secondary machine that's running current Debian Linux, not my daily driver. The improvement suggested with this issue ticket would make SS3 for Linux easier to use and  benefit a variety of use cases.

### Describe alternatives you have considered

I can install ADMB and build my own ss3_linux from source, but it would be more convenient for users in general if the release executables for Linux would be portable and run on all machines.

arni-magnusson:
Thank you Elizabeth, these executables run smoothly on Debian 12. This solves my problem, and it also makes me optimistic that next time when I download the latest SS3 for Linux, it will just run. And that is awesome.

PR #686, opened 2025-07-14
Fix static test new workflow

e-perl-NOAA:

- rewrite .sh file to create portable linux exes
- Add workflow to test build-ss3 artifacts upon success with the simple model
- add step to cancel workflows if multiple are running for the same ref (so as not to waste GHA runners and minutes)

Resolves and closes issue #685 

None.


iantaylor-NOAA (approved):
@e-perl-NOAA, thanks for working on this. The changes look good to me. I'm no expert on shell scripts, but since it solved the problem @arni-magnusson was facing, clearly it worked. Unless you see a reason not to, I would accept the permissions-related change that the github copilot review suggested.

Issue #688, opened 2025-07-21
[Bug]: 3.30.24pre fails Btarget search with EWAA


Rick-Methot-NOAA:
### Issue description

found this while testing Pmack_2023 configuration

### Summary of issue history, discussion, and major aspects of code development

found while testing Pmaxk_2023

### Items to do

- [ ] check code in Btarget search to see if looking for virgin vs benchmark version of target

### Does documentation already exist in the SS3 User Manual?

Yes, the link to the section in the SS3 User Manual is provided below.

### Documentation to add to the SS3 User Manual OR link to existing documentation

nothing yet

### Are r4ss changes needed?

No, this issue doesn't require changes to r4ss

### Are SSI changes needed?

No, this issue doesn't require changes to SSI

PR #692, opened 2025-07-23
Adjust tvg compatibility btarget

Rick-Methot-NOAA:

Adjust usage of the timevary growth compatibility flag such that Btarget will default to the same as it occurs in legacy SS3 before 3.30.24.

* Resolves issue #688 

<-- - [x] No test files are required for this pull request. --I did a local test using the Pmack files from A Jensen>

Check logic of the revised conditional statement.


iantaylor-NOAA (approved):
The changes look OK to me but I'm definitely out of my depth with this section of the code.

iantaylor-NOAA:
Tagging @alexjensen-NOAA so you can see progress on this fix.

e-perl-NOAA:
@Rick-Methot-NOAA I'm working on why the build-warnings is failing. The GHA is behaving differently than running the workflow on codespaces is (which shouldn't be the case) but that is making it hard to debug.

Issue #694, opened 2025-07-28
allow forecast mean recruitment option for late recdevs


Rick-Methot-NOAA:
Implementation of issue #471 caused late recdevs to be disallowed when using forecast recruitment option 4=mult*mean_over_yr_range
It is useful for some users to allow the forecast mean recruitment to be used for late recdevs, but this requires that that mean be calculated during the time series, not at the beginning of the forecast.

This requires creating code like:
if ( forecastmethod==4 && y==lastyearofrange) calculate that mean

PR #695, opened 2025-07-28
improve logic for use of mean forecast recruitment

Rick-Methot-NOAA:

Improve logic controlling the use of mean forecast recruitment to all usage for years after recdev_end

* Resolves issue #694 

A comparison using petrale sole seems helpful

<-- - [X] No, there was no input change. -->

### DIFF
--- a/SS_readcontrol_330.tpl
+++ b/SS_readcontrol_330.tpl
@@ -2374,7 +2374,7 @@
   //  SS_Label_Info_4.6.4 #Setup recruitment deviations and create parm labels for each year
   if (recdev_end > retro_yr)
   {
-    warnstream << " recdev_end: " << recdev_end << " > retro_yr: " << retro_yr << " reset ";
+    warnstream << " last main recdev: " << recdev_end << " to retro_yr: " << retro_yr << " reset ";
     write_message (ADJUST, 0);
     recdev_end = retro_yr;
   }
@@ -2462,7 +2462,7 @@
 
     if (int(Fcast_Loop_Control(3)) >= 4 && Fcast_Rec_yr2 > recdev_end)
   {
-    warnstream << " Error, Fcast_Rec_Yr2 must be <= recdev_end" << Fcast_Rec_yr2 << " > " << recdev_end;
+    warnstream << " last year of averaging for mean forecast recr must be <= last main recdev" << Fcast_Rec_yr2 << " > " << recdev_end;
     write_message(FATAL, 1);
   }
 


SS_readcontrol_330.tpl
```diff
@@ -2459,7 +2459,13 @@
       }
     }
   }
-  
+
+    if (int(Fcast_Loop_Control(3)) >= 4 && Fcast_Rec_yr2 > recdev_end)
+  {
+    warnstream << " Error, Fcast_Rec_Yr2 must be <= recdev_end" << Fcast_Rec_yr2 << " > " << recdev_end;
```
iantaylor-NOAA, line 2465: `Fcast_Rec_Yr2` and `recdev_end` may not make sense to the user (e.g. is `recdev_end` the last recruitment year or the last of the main recdevs? 

Could we change to something like `the range of years to calculate forecast recruitment can't extend beyond the final main recruitment deviation.`? I think we don't need `Error` in the message since `write_message` should add "FATAL".

iantaylor-NOAA (approved):
Looks good. I have one suggested change to the error message but otherwise looks great.

I did look on what happens if you have the deprecated option `Fcast_Loop_Control(3) == 3` and it seems that the setting will be automatically changed to option 4 (with a warning) before it gets to this revised code so makes sense to exclude reference to option 3 as you've done.

Rick-Methot-NOAA:
@e-perl-NOAA 
for section 6.4 of the manual:
currently is:
```
3 = Mean recent recruitment: calculate the mean recruitment and use it during the forecast period.

Note that bias adjustment is not applied to this mean because the values going into the mean have already been bias adjusted.

This feature affects the expected recruitment in all years after the last year of the main recruitment deviations. This means that if the last year of main recruitment deviations is before end year, then the last few recruitments, termed “late”, are also affected by this forecast option. For example, option 3 would allow you to set the last 2 years of the time series and all forecast years to have recruitment equal to the mean recruitment for the last 10 years of the main recruitment era.
```
revise to:
```
3 (deprecated) and 4 = Multiplier times mean recruitment: calculate the mean recruitment from a specified range of years and use it during the forecast and late recdev period.
This mean recruitment is applied to all years after the last year of the main recruitment deviations. Consequently, the last year from which the mean is calculated must be <= the last year for main recruitment deviations. This feature would, for example, allow a user to set the last 2 years of the time series and all forecast years to have recruitment equal to 0.75 * the mean recruitment for the last 10 years of the main recruitment era.
Note that bias adjustment is not applied to this mean because the values going into the mean have already been bias adjusted.
```

Issue #682, opened 2025-06-04
[Feature]: Add warning to avoid use of negative length for the Lmin parameter


iantaylor-NOAA:
### Issue description

The attached model [simple_small_neg_length.zip](https://github.com/user-attachments/files/20599661/simple_small_neg_length.zip) replicates an issue that @gertsevv (Vlada Gertseva) and @shcaba (Jason Cope) ran into for an assessment where the length at age 0 was estimated to be negative (here I fixed L_at_Amin = -30 which caused ages 0-2 to have negative length). The Wt_Beg, Wt_Mid, Len_Mat, and Mat*Fecund values in Biology_at_age_in_endyr all show strange values as illustrated in the plot below based on Mat*Fecund.

I'm not sure what warnings or changes would be helpful, but a few ideas include
* add warning or error if length is negative at any age (you could still have L_at_Amin be negative if Amin is negative)
* add warning or error  if length is negative at any age greater than or equal to First_Mature_Age
* change biology calculations to set maturity and fecundity equal to zero if length is negative.

![Image](https://github.com/user-attachments/assets/08f33605-b90e-4670-8647-c64ef917e3fd)

### Summary of issue history, discussion, and major aspects of code development

This is the first I've come across this.

### Items to do

- [x] Add warning or error
- [ ] Change biology calculations?

The new warning is added in the readcontrol tests.  It is only applied to females because male parameter can legitimately be less than 0 as an offet from females.

looks like this:
Warning 4 Fatal Error! parm init value for Lmin: -10.1 cannot be less than population min length bin 10

### Does documentation already exist in the SS3 User Manual?

Yes, the link to the section in the SS3 User Manual is provided below.

### Documentation to add to the SS3 User Manual OR link to existing documentation

Maybe add text to the description of some of the control file settings for Amin, L_at_Amin, and/or First_Mature_Age.

### Are r4ss changes needed?

No, this issue doesn't require changes to r4ss

### Are SSI changes needed?

No, this issue doesn't require changes to SSI

Rick-Methot-NOAA:
I think the solution is to warn and stop if the L_amin is less than the length at the smallest population bin

PR #698, opened 2025-08-04
add warning for Lmin < len_bin(1)

Rick-Methot-NOAA:

everything is in the issue and title.

* Resolves issue #682 

<-- - [x] Test files are in the issue. -->

None

### DIFF
--- a/SS_readcontrol_330.tpl
+++ b/SS_readcontrol_330.tpl
@@ -1617,6 +1617,18 @@
       mgp_type(Ip, Ip + N_natMparms - 1) = 1; // natmort parms
       Ip += N_natMparms;
       mgp_type(Ip, Ip + N_growparms - 1) = 2; // growth parms
+
+      // check on out of bound Lmin values.  Only check females because males can be offset from females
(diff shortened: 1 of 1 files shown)


Rick-Methot-NOAA:
at least one model is failing because the lower bound for Lmin is set out of bounds:
 -10 45 21.6591 36 10 6 2 0 0 0 0 0 0 0 # L_at_Amin_Fem_GP_1

e-perl-NOAA:
@Rick-Methot-NOAA 10 of our test models are throwing this fatal error. I would suggest that we change these test models so that it doesn't throw the fatal error.

e-perl-NOAA:
@Rick-Methot-NOAA I don't know why run-with-est isn't failing but run-no-est is, that is something that I will try to sort out. Regardless, there are some changes in the artifacts from the run-no-est that you should review.

Rick-Methot-NOAA:
not sure how to check performance of run with -noest.  I will need the ss3.par files.  Can you send me the one for hake_2018

e-perl-NOAA:
Ah, yes that is why there is a different. The run with no est uses the par file and the run with est does not use the par file

PR #713, opened 2025-08-26
turn off extra switches when Do_More_Std == 1

iantaylor-NOAA:
@e-perl-NOAA figured out that the Mac runner for r4ss GitHub actions were failing while Windows and Linux were passing. She was able to export the echoinput which ended with `do SmryBio std labels`. However, I noticed that this line should only occur if the `# extra stdev reporting` was 2 but in this model it was 1. Digging into the SS3 source code revealed that some switches were never set for the case where Do_More_Std == 1, and apparently the Mac filled in the missing value differently than the other operating systems.

@e-perl-NOAA checked this build from this branch to confirm that it resolved the issue.

None.

### DIFF
--- a/SS_readcontrol_330.tpl
+++ b/SS_readcontrol_330.tpl
@@ -1623,8 +1623,6 @@
 
      if ( gp == 1 && WTage_rd == 0)
       {
-      echoinput<<"new check on Lmin: "<<Ip<<endl;
-      echoinput<<MGparm_1(Ip)(1,3)<<endl;
         if (MGparm_1(Ip,1) < len_bins(1))
         {
           warnstream << "parm min for Lmin: " << MGparm_1(Ip,1) << " cannot be less than population min length bin " << len_bins(1);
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA:
@Rick-Methot-NOAA, thanks for looking this over. I just pushed the change to line 5589 as you suggested.
But it sounds like this won't necessarily fix the retro problem.

Rick-Methot-NOAA:
I got i fixed but in looking at code I see that some quantities, like natage_std_yr are reported for endyr.  This is ok even in retro because SS3 goes through all years in retro.  But we could consider whether to provide that se output for retro_yr, rather than endyr

Rick-Methot-NOAA:
echo should be showing mgparm_type for each parm at that stage of the file
read.  I'll investigate as it could be another unitialized variable.

Richard  D. Methot Jr.

Stock Assessment Research Scientist (ST)

Northwest Fisheries Science Center

NOAA Fisheries | U.S. Department of Commerce

Office: (425) 666-9893

Mobile: (301) 830-2454

www.fisheries.noaa.gov

Rick-Methot-NOAA (approved):
fixed issue

PR #732, opened 2025-09-16
Change issue linking format in PR template

e-perl-NOAA:

* Resolves issue #


iantaylor-NOAA (approved):
I manually edited the template when creating PR #734 and it successfully closed #733 automatically, so I'm now confident that this change should work (or maybe it didn't work for some other reason, but there's definitely no harm in making this change).

PR #752, opened 2026-02-04
new approach for weight composition - superceded

### LINKED ISSUE #751: need warning and improved logic for generalized sizecomp
A user configured a generalized sizecomp with narrow weight bins for the data.  SS3 crashed without providing warning.  The problem is that when the setup is using weight bins, a narrow population length bin can translate to weight range that is wider than the user-specified weight bins.  
@N-DucharmeBarth-NOAA

The code is designed to detect where the population bin is completely within a data bin and assign all of that pop bin to the data bin, and to detect when there are multiple consecutive pop bins that fall within the same data bin range and assign all of them to that range.  It also can detect when a pop bin stradles the upper edge of a data bin and apportion that pop bin to the current and nect data bin, then to correctly move on to the next pop bin.
However, the code cannot detect when the upper edge of a pop bin would skip over a data bin.  When this happens, no warning is issued and a crash typically occurs.

- [ ] minimum fix would be to provide a fatal warning in all such circumstances
- [ ] best fix would be to develop an algorithm by which the code would apportion that wide population bin to a range of data bins.  Surely such an algorithm has already been developed somewhere and we just need to know what to ask for.

Rick-Methot-NOAA:

* Resolves #751 

## What tests have been done? 
in progress

### DIFF
--- a/.github/workflows/test-simple-with-ss3-artifacts.yml
+++ b/.github/workflows/test-simple-with-ss3-artifacts.yml
@@ -107,7 +107,7 @@ jobs:
           fi
 
       - name: Archive Simple model output
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: simple-model-output-${{ matrix.os }}
           path: ss3-test-models/models/Simple/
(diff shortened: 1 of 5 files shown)


Rick-Methot-NOAA:
This PR is confusing with the misdirected copilot comments.  Will close and restart a new PR.

PR #773, opened 2026-07-22
Bug: Fix VBK sign for Nat_M case 6

nathanvaughan-NOAA:

This is a fix for the VBK sign change in natural mortality option 6 

* This resolves the same bug that was corrected for natural mortality option 2 in commit 1609ef

 [x] No test files are required for this pull request. 

Presumably the required tests for this change were already completed when the fix was made to case 2

### DIFF
--- a/SS_biofxn.tpl
+++ b/SS_biofxn.tpl
@@ -1229,7 +1229,7 @@ FUNCTION void get_natmort()
             //  SS_Label_Info_17.1.2.6  #case 6:  Calculate lorenzen M from survivorship over fixed age range
             case 6: //  Survivorship based Lorenzen M
             {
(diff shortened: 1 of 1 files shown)


Rick-Methot-NOAA:
Good catch Nathan.  I'll approve after the build failure is addressed.
@e-perl-NOAA

e-perl-NOAA:
@Rick-Methot-NOAA and @nathanvaughan-NOAA, there is no reason that the windows build should fail, especially for the error that it's giving. I made the same small change in a branch and it [built on all OS](https://github.com/nmfs-ost/ss3-source-code/actions/runs/29954449555). I think we can chalk this up to a github actions quirk and merge it in.

PR #780, opened 2026-07-27
repair density-dependent parameter functionality

### LINKED ISSUE #779: [Bug]: error in invoking density-dependence
### Describe the bug

parent issue #778 included an error in () for setting up density-dependent parameters.
A test with the current code (3.30.25) confirmed that density-dependence was not happening for a MGparm.
With this control.ss line:
```
# Sex: 1  BioPattern: 1  Growth
 -10 45 21.6591 36 10 6 2 -203 0 0 0 0 0 0 # L_at_Amin_Fem_GP_1
```

The echoinput showed:
```
timevary for MG parameter: 2
baseparm:  -10 45 21.6591 36 10 6 2 -203 0 0 0 0 0 0
env link_type: 2 env_var: 0
 do env additive 
timevary_setup 1 2 1 0 0 2 0 0 0 0 0 0 2 0
Display timevary_MG flag
1970 timevary_MG:  0 0 0 0 0 0 0 0
1971 timevary_MG:  0 0 0 0 0 0 0 0
```

And the parameter stayed constant over time.

### To Reproduce

set as shown above

### Expected behavior

quantity calculated from the parameter should have been time varying.

### Screenshots

_No response_

_No response_

3.30.25 and probably several earlier versions

### Additional Context

_No response_

Rick-Methot-NOAA:

* Resolves #779

### DIFF
--- a/SS_global.tpl
+++ b/SS_global.tpl
@@ -95,6 +95,7 @@ GLOBALS_SECTION
   adstring onenum2(4);
   adstring anystring;
   adstring anystring2;
+  adstring DD_name;
   adstring report_sso_filename;
   adstring MSY_name; // label describing what Do_MSY and MSY_units are being used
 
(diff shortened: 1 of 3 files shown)


iantaylor-NOAA (approved):
Looks good for getting things working again.

I haven't used the density dependence much, so set up the attached model [simple_small_DD.zip](https://github.com/user-attachments/files/30468207/simple_small_DD.zip) to include both a density dependent L_at_Amin and an environmental variable link to L_at_Amax which ended up with parameter labels and `L_at_Amin_Fem_GP_1_ENV_add_DD` and `L_at_Amax_Fem_GP_1_ENV_addX`. 

This made me wonder about a few more trivial changes:
1. Change from "X" to an empty string "" for the standard environmental link parameter.
2. It looks like the ENVIRONMENTAL_DATA output is not included in Report.sso unless there is environmental data in the data file. However, perhaps that table should be reported in cases where a DD link is used but no environmental data is provided in order for users to have access to the columns of processed time series data. 
3. It looks like this text in the user manual (here: https://github.com/nmfs-ost/ss3-doc/blob/main/13output.tex#L129) needs to be revised to note that there is both summary numbers and summary biomass:
  > Density-dependence can be used by linking to population quantities that have already been calculated at the start of the year. These include summary biomass, spawning biomass, and recruitment deviations. These three quantities are mapped into the -1, -2, and -3 columns of the environmental data matrix where they can be used as if there were environmental data input.

Rick-Methot-NOAA:
good suggestions.  The "X" was just leftover from debugging and will get removed.

